PR #4, opened 2022-09-26
PLAT-5622: Integrate with the deployer

fraenkel:
Expose what is necessary for the deployer.
    IAM performed by respective modules.
    Cleanup paths


main.tf
```diff
@@ -130,24 +115,41 @@ module "eks" {
...
+  subnets = [for s in local.private_subnets : {
+    name       = s.name
+    id         = s.id
+    cidr_block = s.cidr_block
+  }]
+  vpc_id = module.network.vpc_id
+  roles  = module.eks.eks_node_roles
+}
+
+data "aws_iam_role" "eks_master_roles" {
```
miguelhar, line 139: Shouldnt this live in `submodules/eks`?

main.tf
```diff
@@ -130,24 +115,41 @@ module "eks" {
...
-  k8s_cluster_endpoint    = module.eks.cluster_endpoint
-  managed_nodes_role_arns = module.eks.managed_nodes_role_arns
-  eks_master_role_names   = concat(var.eks_master_role_names, module.eks.eks_master_role_name)
-  kubeconfig_path         = local.kubeconfig_path
+  source               = "./submodules/k8s"
+  ssh_pvt_key_path     = abspath(local.ssh_pvt_key_path)
+  bastion_user         = local.bastion_user
+  bastion_public_ip    = try(module.bastion[0].public_ip, "")
+  k8s_cluster_endpoint = module.eks.cluster_endpoint
+  eks_node_role_arns   = [for r in module.eks.eks_node_roles : r.arn]
```
miguelhar, line 150: Either the var is not named correctly or its contents do not match,  i would expect this var to produce
```hcl
managed_nodes_role_arns = module.eks.managed_nodes_role_arns
```

outputs.tf
```diff
@@ -13,37 +13,31 @@ output "hostname" {
...
+  value       = module.storage.efs_file_system
 }
 
 output "s3_buckets" {
-  description = "S3 buckets name,arn."
+  description = "S3 buckets"
   value       = module.storage.s3_buckets
 }
 
+output "domino_key_pair" {
```
miguelhar, line 31: Do we need this?

miguelhar (approved):
Changes are cosmetic, Approving for functinality

main.tf
fraenkel: I don't want to make hard assumptions about what the upper layer require.
You originally return the names and then converted it back to arns.
Yes, I could return arns but it gets very messy to determine what is needed by other modules so let them decide the inputs and make the outputs pretty transparent.

outputs.tf
fraenkel: got rid of yours since we only needed one.

main.tf
fraenkel: you didn't pass the eks_master_roles to eks so I just left things alone.  You have the user provide additional role names but we need arns so we need to convert them somewhere. We already have access to arns and names from what eks creates.
Was this supposed to do something else?

PR #5, opened 2022-09-28
PLAT-5622: Simplify the variables

fraenkel:


submodules/eks/README.md
```diff
@@ -13,8 +13,8 @@
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 4.0 |
-| <a name="provider_null"></a> [null](#provider\_null) | >= 3.1.0 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | 4.32.0 |
```
miguelhar, line 16: How did this get pinned ?

fraenkel: it gets generated based on whatever it evals to at generation which is odd but how it works.

miguelhar (approved):

submodules/eks/README.md
miguelhar: it should look at the provider constraint which is `>= 4.0` and not pin to a single version

PR #6, opened 2022-10-03
PLAT-5622 k8s provision through socks proxy

miguelhar:


submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -34,42 +22,49 @@ check_kubeconfig() {
...
   echo
 }
 
+validate_url() {
+  local url="$1"
+  if curl --head --silent --fail -o /dev/null "$url"; then
+    return 0
+  else
+    return 1
+  fi
```
fraenkel, line 52: why not send back the rc ($?) ?

miguelhar: We can, since any non-zero will work, Preferred an explicit conditional return

submodules/eks/cluster.tf
```diff
@@ -119,8 +119,13 @@ resource "null_resource" "kubeconfig" {
   provisioner "local-exec" {
     command = "aws eks update-kubeconfig --kubeconfig ${var.kubeconfig_path} --region ${var.region} --name ${aws_eks_cluster.this.name}"
   }
+  provisioner "local-exec" {
+    when    = destroy
+    command = "rm -f ${self.triggers.kubeconfig_file}"
```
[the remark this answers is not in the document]
miguelhar: that is true. Wanted to address stale kubeconfigs.

submodules/k8s/templates/k8s-functions.sh.tftpl
fraenkel: I just realized since this is silent, do we need to say something on error.

submodules/eks/cluster.tf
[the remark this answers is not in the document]
fraenkel: I would think its only safe to delete if we knew we created it. Not sure I want to track that bit of detail.
could do a `k config delete-context`

submodules/k8s/templates/k8s-functions.sh.tftpl
miguelhar: its covered by the client call 
```
  if test -f "$k8s_manifest" || validate_url "$k8s_manifest"; then
    echo "Applying $k8s_manifest..."
    (
      export HTTPS_PROXY=socks5://127.0.0.1:${k8s_tunnel_port} && kubectl apply -f $k8s_manifest
    )
  else
    printf "$RED $k8s_manifest does not exist. $EC \n" && exit 1
  fi
```

fraenkel: no, that is the first half, not the validate_url which we have no real idea why it failed.

fraenkel (approved):

PR #8, opened 2022-10-26
PLAT-3406: initial CI

fraenkel:
verify that the infrastructure can be created and no drift occurs afterward

### DIFF
--- a/tests/main.tf
+++ b/tests/main.tf
@@ -1,10 +1,10 @@
 module "domino_eks" {
-  source                       = "../../terraform-aws-eks"
+  source                       = "./.."
   deploy_id                    = var.deploy_id
   region                       = var.region
   number_of_azs                = 2
   k8s_version                  = var.k8s_version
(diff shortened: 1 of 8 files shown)


tests/main.tf
```diff
@@ -1,10 +1,10 @@
 module "domino_eks" {
-  source                       = "../../terraform-aws-eks"
+  source                       = "./.."
   deploy_id                    = var.deploy_id
   region                       = var.region
   number_of_azs                = 2
   k8s_version                  = var.k8s_version
-  route53_hosted_zone_name     = "infra-team-sandbox.domino.tech"
+  route53_hosted_zone_name     = "deploys-delta.domino.tech"
```
miguelhar, line 7: is this intentional?

miguelhar (approved):
Minor comment

tests/main.tf
fraenkel: yes, its where we do the test deploys

PR #12, opened 2022-11-22
PLAT-3795: IAM policies for deployment

Secretions:


variables.tf
```diff
@@ -16,6 +16,24 @@ variable "deploy_id" {
...
+  type        = string
+  description = "Explicit AWS access key id to use"
+  default     = ""
+}
+
+variable "secret_key" {
+  type        = string
+  description = "Explicit AWS secret access key id to use"
+  default     = ""
+}
```
fraenkel, line 29: why would we want them to provide this if it can be obtained via the environment?

submodules/eks/cluster.tf
```diff
@@ -115,10 +115,35 @@ resource "aws_eks_addon" "this" {
   ]
 }
 
+resource "null_resource" "kubeconfig-no-role" {
```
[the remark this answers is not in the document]
Secretions: Just to be clear, you're just saying consolidate?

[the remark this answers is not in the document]
fraenkel: if possible

iam-bootstrap/iam.tf
```diff
@@ -0,0 +1,37 @@
...
+          AWS = "arn:aws:iam::${data.aws_caller_identity.admin.account_id}:root"
+        }
+      },
+    ]
+  })
+
+  managed_policy_arns = aws_iam_policy.deployment[*].arn
+
+  provisioner "local-exec" {
+    command = "sleep 15"
```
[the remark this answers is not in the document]
Secretions: I can probably nuke this. Believe I needed this in the original version of this where it was all running in a single terraform pass.

variables.tf
Secretions: Although I put it there when I was considering a user rather than a role, I left it thinking perhaps someone (not using the deployer) might want prefer it specifically so it can't get it from the environment. As in, to better enforce that it's using whatever user they've created attached to their deployment policy.

Similarly, you could actually say the same thing about the role. Because of the way I've automated the platform apps side, it makes use of it, but we could conceivably just assume the role prior to running terraform and keep the terraform a bit simpler. Basically reduces this PR to just the iam policy portion, at the expense of a platform-apps-side awscli call to sts some minor environment variable management.

fraenkel: So there are multiple issues. What you have may be insufficient like with our environments, we have a session token. If you look at the laundry list of options, its almost always best to just define this stuff outside either through env vars or AWS config. If you are running on an ec2 instance, you can pick up details from the metadata. Point is this is tricky to get right but easy to be insufficient or a security issue.  Most people store their tfvars, but no one would store it with secrets. This means you have to pass it on the command line or env vars which gets you back to where we started.

Secretions: I'm sold on deleting the key/secrets. What do you think about the role part?

fraenkel: Given we assume the role in other parts, it makes sense to do it as part of the provider too.

fraenkel: I am curious. if you configured the AWS CLI to do assume role, and you don't pass any assume_role_arn in, does it succeed?

Secretions: Assuming the role in other parts would go away too. It'd be something like:

* Remove all changes in this PR outside of the `iam-bootstrap` subdirectory
* Run the `aws sts assume-role etc....` in the deployer, and put the key/secret/token in the environment on the localrunner that runs terraform
* Document for users that its their responsibility to make sure they're running with their policies
  * Maybe also make an option to create a user vs role to the iam tf, since that might be more convenient at that point.

I don't see any reason this wouldn't do what's desired. The only thing in the environment for that terraform instance and all its child processes becomes the assumed role's key/secret/token, and we no longer have to do anything different within the module itself.

fraenkel: so don't assume calling assume-role will do much of anything.
we can't do it since we in most cases already get an assumed role from vault.

For others, they may have different integrations that give them creds since most folks don't have personal creds.

fraenkel: I mispoke. There are ways to make this work.

fraenkel: I am just wondering, if I already had to set up some of my creds before invoking the deployer, why not make me set it up exactly as I need before invoking the deployer.

Secretions: That's basically what I'm saying here. The deployer support that does it in one go is entirely for the purposes of automating testing this in CI. Since that's the case, customers using this module directly can just run the iam part separately, set up their environment to use it however they want and just run the module without any special configuration. Likewise, in the real world, if one of our deploy engineers is using the deployer, they might not use the built-in support for the IAM and just set it up separately.

The above was already the idea behind all of this, so the assume_role_arn had just two intents behind it. First, a way to run it in the deployer CI so we're testing the policies. Second, the hypothetical "customer wants to audit that they're using the role by virtue of the fact that it's in their config" thing. Which maybe isn't worth worrying about.

Let me just make an update to both PRs of this other version that I'm thinking of so we avoid talking past each other.

PR #15, opened 2022-12-06
PLAT-5933: Support govcloud

fraenkel:
Use appropriate AWS partition when generating ARNs.


steved (commented):
Is there a tflint check or something that could catch this?

fraenkel:
added tflint bits

steved (approved):
Thanks!

PR #16, opened 2022-12-19
PLAT-3795: Terraform module and IAM policies for deployment

Secretions:
* Policies meant to be deployed and used separate from the main module
* End users / deploy engineers would provision the module (or manually provision the policies), manually assume the role (or use an associated user they've created) and then provision the main EKS module
* The main EKS module doesn't have any interaction with the policy, as it does not need to know about it
* See https://github.com/cerebrotech/platform-apps/pull/6505 for the internal deployer automation for this

### DIFF
--- a/iam-bootstrap/iam.tf
+++ b/iam-bootstrap/iam.tf
@@ -10,7 +10,18 @@ resource "aws_iam_policy" "deployment" {
 
   name = "${var.deploy_id}-deployment-policy-${count.index}"
 
-  policy = templatefile(abspath(pathexpand(local.iam_policy_paths[count.index])), merge({account_id=data.aws_caller_identity.admin.account_id, deploy_id=var.deploy_id, region=var.region, partition=data.aws_partition.current.partition},var.template_config))
+  policy = templatefile(
+    abspath(pathexpand(local.iam_policy_paths[count.index])),
+    merge(
+      {
+        account_id = data.aws_caller_identity.admin.account_id,
+        deploy_id  = var.deploy_id,
+        region     = var.region,
+        partition  = data.aws_partition.current.partition
+      },
+      var.template_config
+    )
+  )
 }
 
 resource "aws_iam_role" "deployment" {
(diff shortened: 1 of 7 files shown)


iam-bootstrap/iam.tf
```diff
@@ -0,0 +1,36 @@
...
+
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Sid    = ""
+        Principal = {
+          AWS = "arn:aws:iam::${data.aws_caller_identity.admin.account_id}:root"
```
fraenkel, line 27: govcloud

iam-bootstrap/render.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+export deploy_id=example
```
fraenkel, line 3: why do we have this file?

iam-bootstrap/iam.tf
```diff
@@ -0,0 +1,36 @@
...
+
+data "aws_caller_identity" "admin" {}
+data "aws_partition" "current" {}
+
+resource "aws_iam_policy" "deployment" {
+  count = length(local.iam_policy_paths)
+
+  name = "${var.deploy_id}-deployment-policy-${count.index}"
+
+  policy = templatefile(abspath(pathexpand(local.iam_policy_paths[count.index])), merge({account_id=data.aws_caller_identity.admin.account_id, deploy_id=var.deploy_id, region=var.region, partition=data.aws_partition.current.partition},var.template_config))
```
steved, line 13: can this be a bit more readable? either separate lines or using more locals?

iam-bootstrap/iam.tf
steved: Seems like a miss on the tflint side as it doesn't check this, but should we be running pre-commit in CI to (theoretically) catch this stuff?

steved (approved):

iam-bootstrap/iam.tf
Secretions: Fixed this instance, anyhow.

iam-bootstrap/render.sh
Secretions: A quick helper for anyone using the policies, but not provisioning them via the terraform  module.

iam-bootstrap/iam.tf
Secretions: Sure, I made it multiline.

PR #18, opened 2023-01-03
PLAT-5668: Compatibility fixes for coming from CDK

Secretions:
Just a few misc. things that needed to change to comfortable import EKS infra created by our CDK app.

### DIFF
--- a/iam.tf
+++ b/iam.tf
@@ -27,13 +27,13 @@ data "aws_iam_policy_document" "route53" {
 
 resource "aws_iam_policy" "route53" {
   count  = var.route53_hosted_zone_name != "" ? 1 : 0
-  name   = "${var.deploy_id}-Route53"
+  name   = "${var.deploy_id}-route53"
   path   = "/"
   policy = data.aws_iam_policy_document.route53[0].json
 }
 
 resource "aws_iam_role_policy_attachment" "route53" {
-  for_each   = var.route53_hosted_zone_name != "" ? toset([for r in module.eks.eks_node_roles : r.name]) : []
+  count      = var.route53_hosted_zone_name != "" ? length(module.eks.eks_node_roles) : 0
   policy_arn = aws_iam_policy.route53[0].arn
-  role       = each.value
+  role       = lookup(module.eks.eks_node_roles[count.index], "name")
 }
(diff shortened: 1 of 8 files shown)


iam.tf
```diff
@@ -27,13 +27,13 @@ data "aws_iam_policy_document" "route53" {
...
   path   = "/"
   policy = data.aws_iam_policy_document.route53[0].json
 }
 
 resource "aws_iam_role_policy_attachment" "route53" {
-  for_each   = var.route53_hosted_zone_name != "" ? toset([for r in module.eks.eks_node_roles : r.name]) : []
+  count      = var.route53_hosted_zone_name != "" ? length(module.eks.eks_node_roles) : 0
   policy_arn = aws_iam_policy.route53[0].arn
-  role       = each.value
+  role       = lookup(module.eks.eks_node_roles[count.index], "name")
```
Secretions, line 38: This was to fix the fact that terraform refresh can fail here under some circumstances.

submodules/storage/s3.tf
```diff
@@ -414,7 +414,7 @@ resource "aws_s3_bucket_request_payment_configuration" "buckets_payer" {
 }
 
 resource "aws_s3_bucket_logging" "buckets_logging" {
-  for_each      = { for k, v in local.s3_buckets : k => v if v.bucket_name != aws_s3_bucket.monitoring.bucket }
+  for_each      = { for k, v in local.s3_buckets : k => v if k != "monitoring" }
```
Secretions, line 417: This was also to fix the fact that terraform refresh can fail here under some circumstances.

iam.tf
steved: Just curious, what does this change that allows it to succeed? Or, what fails during refresh?

steved (approved):
Changes look fine, just curious about the refresh failures :+1:

iam.tf
Secretions: This is the error:

```
The "for_each" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the for_each depends on.
```

@miguelhar said that count fixes it and gave me the snippet, and I was able to confirm it did. These failures happen when things are partially provisioned and you're retrying, which I just happened to run into while figuring out what permissions were missing (thing provisions -> fix iam -> can't continue because refresh fails).

PR #19, opened 2023-01-05
PLAT-6023: Add support for optional pod subnet

Secretions:
* Adds optional pod subnet option
* Sequencing changes so this works out of the box
  * Moves `k8s` submodule invocation into eks module (chained)
    * So we can set this up before the nodegroups get created
    * Otherwise, requires cycling any extant nodes
  * Always install vpc cni module
    * Needs to be setup after vpc cni addon installation
  * Install remaining addons afterward like normal
    * At least one of the other addons can fail without nodes


main.tf
```diff
@@ -89,6 +94,12 @@ locals {
   public_cidr_blocks = slice(local.subnet_cidr_blocks, 0, local.num_of_azs)
   ## Match the private subnet var to the list of cidr blocks
   private_cidr_blocks = slice(local.subnet_cidr_blocks, local.num_of_azs, length(local.subnet_cidr_blocks))
+  ## Determine cidr blocks for pod network
+  base_pod_cidr_network_bits = tonumber(regex("[^/]*$", var.pod_cidr))
+  pod_cidr_blocks = var.use_pod_cidr == false ? [] : cidrsubnets(
```
steved, line 99: nit `!var.use_pod_cidr ?`?

submodules/network/variables.tf
```diff
@@ -48,6 +53,25 @@ variable "cidr" {
...
+    )
+    error_message = "Argument cidr must be a valid CIDR block."
+  }
+}
+
+variable "use_pod_cidr" {
+  type        = bool
+  description = "Use additional pod CIDR range (ie 100.64.0.0/16) for pod/service networking"
+  default     = true
+}
```
steved, line 73: More a philosophical question than anything, why two vars vs. an `optional(string)` pod_cidr where null means it's disabled?

submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -32,6 +32,22 @@ set_k8s_auth() {
...
+  fi
+  kubectl_cmd -n kube-system set env daemonset aws-node AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG=true
+  if test -f "$ENICONFIG_YAML"; then
+    printf "$GREEN Updating $ENICONFIG_YAML... $EC \n"
+    kubectl_apply "$ENICONFIG_YAML"
+  else
+    printf "$RED $ENICONFIG_YAML does not exist. $EC \n" && exit 1
+  fi
+  echo
+  kubectl_cmd -n kube-system set env daemonset aws-node ENI_CONFIG_LABEL_DEF=topology.kubernetes.io/zone
```
steved, line 48: Just from a safety-ish perspective, why not apply the aws-node changes once since `set env` takes multiple env variables? Would it also be worth `kubectl wait`'ing to ensure it rolls out successfully and isn't left crashlooping?

Secretions: For the first question, just dogmatically doing things in the same order as the aws docs. There's probably no good reason for it. For the second, there's nothing to roll out since there are no nodes.

submodules/network/variables.tf
Secretions: I actually started doing it that way, which looks something like this: https://github.com/dominodatalab/terraform-aws-eks/pull/22/files

However, I was a bit sketched out by what I had to do to the validation block:

```
    condition = (
      var.pod_cidr == null ||
      (try(cidrhost(var.pod_cidr, 0), null) == try(regex("^(.*)/", var.pod_cidr)[0], null) &&
      try(cidrnetmask(var.pod_cidr), null) == "255.255.0.0")
    )
```

Specifically, having to put the try on the right side as well. There were a couple minor things I didn't like (`pod_cidr` then works differently than `cidr`, additional exception to local variable calculation) that weren't a big deal, but that kind of made me decide not to go that route.

Secretions: I should also add that I kind of expect that the most likely user experience is toggling it on and off anyway, since there's no good reason to customize away from the carrier grade nat, so I'm sort of conceiving of the boolean as the standard user-facing variable and the other variable as the, uhh, "expert mode" variable.

submodules/k8s/templates/k8s-functions.sh.tftpl
Secretions: Updated this to a single call.

main.tf
Secretions: Your other question made me notice I'm checking for `use_pod_cidr != null` in one part too. Both of these changed.

submodules/k8s/templates/k8s-functions.sh.tftpl
steved: Oh yeah... duh. Thanks.

steved (approved):

Secretions:
I don't believe the circle jobs not kicking off could be related to anything I've done here. I'm going to admin merge.

PR #20, opened 2023-01-06
Miguelhar.plat 3687.mod upgrade test

miguelhar:
[PLAT-3687](https://dominodatalab.atlassian.net/browse/PLAT-3687)

- Passive terraform checks run in GHA on push.
- GHA triggers `CircleCi` deploy/upgrade tests on pull request.
- Terraform upgrade verifies the ability to successfully apply . Follow up, consider what kind of changes are "allowed".
- Converted `main.tf` to `main.tf.json` for simplistic(`jq`) module source upgrades.
- Upgrades CI terraform version to `1.3.7`.
- Set `tflint` `deep_check` to `false`.
- Adds `tfsec` to pre-commits.


.circleci/config.yml
```diff
@@ -75,17 +68,66 @@ jobs:
             terraform workspace select default
             terraform workspace delete ${WORKSPACE}
 
-parameters:
-  terraform_version:
-    default: "1.3.6"
-    type: string
+  test-upgrade:
+    docker:
+      - image: cimg/aws:2022.09
```
ddl-dclegg, line 73: @miguelhar Have we (or can we) try testing this with the [latest](https://circleci.com/developer/images/image/cimg/aws) (`2022.11.1`)? If it causes issues, I'm okay with staying on `2022.09`

ddl-dclegg (approved):
I think this looks good. Once all the checks are passing, I'm all 👍

.circleci/config.yml
```diff
@@ -61,31 +54,75 @@ jobs:
...
+      - run: &terraform_destroy
           name: Terraform destroy
           working_directory: tests
           command: |
-            terraform destroy -auto-approve
+            terraform destroy --auto-approve || terraform destroy --auto-approve --refresh=false
           when: always
 
-      - run:
+      - run: &terraform_ws_delete
```
steved, line 64: Has nothing to do with your changes, but if this doesn't run `always` so won't be collecting PR workspaces over time if the workflow errors? Is that ok?

.circleci/config.yml
```diff
@@ -61,31 +54,75 @@ jobs:
             # we will always have drift
             # terraform plan -detailed-exitcode
 
-      - run:
+      - run: &terraform_destroy
```
steved, line 57: FWIW instead of anchors, you can now share "commands": https://github.com/dominodatalab/terraform-aws-eks/pull/21

.github/workflows/deploy-upgrade.yml
```diff
@@ -0,0 +1,21 @@
...
+    types: [opened, synchronize]
+jobs:
+  test-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Test-Deploy
+        id: test-deploy-workflow
+        uses: CircleCI-Public/trigger-circleci-pipeline-action@v1.0.6
+        env:
+          CCI_TOKEN: ${{ secrets.CIRCLE_PERSONAL_TOKEN }}
```
steved, line 13: let's switch this to domino-builds before merging (contact devprod)

.circleci/config.yml
miguelhar: I dont think it matters for this one, since its using local state, but Ill add the always for consistency.

.circleci/config.yml
```diff
@@ -1,91 +1,145 @@
...
+    environment:
+      WORKSPACE: << parameters.workspace >>
+    steps:
+      - checkout
+      - install_tf
+      - set_env_vars
+      - gen_pvt_key
+      - tf_init_apply
+      - tf_destroy
+      - tf_ws_delete
```
fraenkel, line 91: shouldn't these have a when: always?

fraenkel (commented):
I was trying to avoid managing two pre-commits, one in the GHA and one in pre-commit.yaml.
Its a bit odd that GHA checks all but locally we only check what changed.

.circleci/config.yml
miguelhar: [here](https://github.com/dominodatalab/terraform-aws-eks/blob/235b3b573b300de75b9292b5f95b3272e5f0011e/.circleci/config.yml#L60) and [here](https://github.com/dominodatalab/terraform-aws-eks/blob/235b3b573b300de75b9292b5f95b3272e5f0011e/.circleci/config.yml#L70)

.circleci/config.yml
```diff
@@ -1,91 +1,145 @@
...
+          command: |
+            echo "Testing terraform module upgrade"
+            export MOD_SOURCE="./.."
+            echo "Setting module source to local ref: ${MOD_SOURCE}"
+            cat \<<< $(jq --arg mod_source "${MOD_SOURCE}" '.module[0].domino_eks.source = $mod_source' main.tf.json) >main.tf.json
+            terraform init --reconfigure --upgrade
+            terraform validate
+            terraform apply --auto-approve
+      - tf_destroy
+      - tf_ws_delete
```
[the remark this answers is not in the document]
miguelhar: [here](https://github.com/dominodatalab/terraform-aws-eks/blob/235b3b573b300de75b9292b5f95b3272e5f0011e/.circleci/config.yml#L70)

miguelhar:
> I was trying to avoid managing two pre-commits, one in the GHA and one in pre-commit.yaml. Its a bit odd that GHA checks all but locally we only check what changed.

The pre-commit binaries will need to be installed regardless of ci. We can always remove all the actions and just have a single `uses: pre-commit/action@v3.0.0` by default it will run all the pre-commits. It was done this way to improve performance, since they all run in parallel with the flexibility to turn/off on.

.circleci/config.yml
fraenkel: what is unclear is whether a command as a job step is implicity replace or just treated as a separate sequence of steps. In which case the job steps above would need to denote that a given command will always execute and most likely the steps within the command as well.
The CircleCI docs are unclear on the exact semantics. But if you have seen errors from any previous step cause destroy and delete, I am ok with it.

fraenkel:
We can always flip things via the env variable SKIP_CHECKS.
The bigger issue is always keeping multiple sources in sync and running them serial doesn't take long in the whole scheme of things. The deploy is going to take the bulk anyhow so whether you fail in 30s or 1m is immaterial to the 30+ minutes.

miguelhar:
> We can always flip things via the env variable SKIP_CHECKS. The bigger issue is always keeping multiple sources in sync and running them serial doesn't take long in the whole scheme of things. The deploy is going to take the bulk anyhow so whether you fail in 30s or 1m is immaterial to the 30+ minutes.

Until a PR is opened the deploy/upgrade will not run which not only save necessary runs but you get the pre-commit validations. 

What is precisely the piece that needs to be kept in sync? and how does Circle mitigates it?

fraenkel:
The actual list of pre-commit hooks. Before pre-commit had X hooks and GHA was doing Y.
If we change anything in pre-commit.yaml we also need to potentially update GHA.

.circleci/config.yml
fraenkel: The docs claim the command steps are inserted so what you have will work.

miguelhar:
> The actual list of pre-commit hooks. Before pre-commit had X hooks and GHA was doing Y. If we change anything in pre-commit.yaml we also need to potentially update GHA.

I would not say anything, but only if we change the hook id, which is unlikely to change.

fraenkel:
Just telling how it all ended up when I switched to just run pre-commit run --all.  We had a different set of things executed with different args. Maybe a comment at the top to make sure people check both.

steved (approved):
:+1: once any other concerns are addressed

miguelhar:
> Just telling how it all ended up when I switched to just run pre-commit run --all. We had a different set of things executed with different args. Maybe a comment at the top to make sure people check both.

Note added to pre-commits

.github/workflows/deploy-upgrade.yml
```diff
@@ -0,0 +1,21 @@
+name: Trigger Deploy-Upgrade
+on:
+  pull_request:
+    types: [opened, synchronize]
```
fraenkel, line 4: reopened?
or just use the default which is all 3.

fraenkel (approved):
should also modify dependabot for GHA

fraenkel (approved):

PR #22, opened 2023-01-10
Put back pod_cidr as trigger

Secretions:


Secretions:
Only made this as an example, closing.

PR #25, opened 2023-01-17
Set proxy-url in kubeconfig (temporarily)

Secretions:
SOCKS proxy is incompatible with awscli. This normally doesn't matter the way this module is used, but when a role is specified then awscli needs to call out and fails at the socks proxy.

By moving the proxy config into the kubeconfig, kubectl uses the proxy and awscli doesn't, fixing this issue.

I use `sed` to add the `proxy-url` key rather than `yq`, even though it's dumber, to avoid adding an additional pre-req.

### DIFF
--- a/submodules/k8s/templates/k8s-functions.sh.tftpl
+++ b/submodules/k8s/templates/k8s-functions.sh.tftpl
@@ -17,7 +17,8 @@ check_kubeconfig() {
   printf "$GREEN Checking if $KUBECONFIG exists... $EC \n"
   if test -f "$KUBECONFIG"; then
     echo "$KUBECONFIG exists, creating $KUBECONFIG_PROXY for proxy use."
-    sed -e '/server:/a\' -e "    proxy-url: socks5://127.0.0.1:${k8s_tunnel_port}" $KUBECONFIG > $KUBECONFIG_PROXY
+    cp $KUBECONFIG $KUBECONFIG_PROXY
+    kubectl --kubeconfig $KUBECONFIG_PROXY config set "clusters.${eks_cluster_arn}.proxy-url" "socks5://127.0.0.1:${k8s_tunnel_port}"
   else
     echo "$KUBECONFIG does not exist." && exit 1
   fi
--- a/submodules/k8s/templates/k8s-functions.sh.tftpl
+++ b/submodules/k8s/templates/k8s-functions.sh.tftpl
@@ -17,7 +17,7 @@ check_kubeconfig() {
   printf "$GREEN Checking if $KUBECONFIG exists... $EC \n"
   if test -f "$KUBECONFIG"; then
     echo "$KUBECONFIG exists, creating $KUBECONFIG_PROXY for proxy use."
-    sed -e '/server:/a\' -e '    proxy-url: socks5://127.0.0.1:1080' $KUBECONFIG > $KUBECONFIG_PROXY
+    sed -e '/server:/a\' -e "    proxy-url: socks5://127.0.0.1:${k8s_tunnel_port}" $KUBECONFIG > $KUBECONFIG_PROXY
   else
     echo "$KUBECONFIG does not exist." && exit 1
   fi
--- a/submodules/k8s/templates/k8s-functions.sh.tftpl
+++ b/submodules/k8s/templates/k8s-functions.sh.tftpl
@@ -4,6 +4,9 @@ RED="\e[31m"
 GREEN="\e[32m"
 EC="\e[0m"
 
+KUBECONFIG="${kubeconfig_path}"
+KUBECONFIG_PROXY="$KUBECONFIG-proxy"
+
 open_ssh_tunnel_to_k8s_api() {
   TUNNEL_SOCKET_FILE=$${TUNNEL_SOCKET_FILE:-/tmp/k8s-tunnel-socket}
   printf "$GREEN Openning k8s tunnel ... $EC \n"
@@ -11,10 +14,10 @@ open_ssh_tunnel_to_k8s_api() {
 }
 
 check_kubeconfig() {
-  local KUBECONFIG="${kubeconfig_path}"
   printf "$GREEN Checking if $KUBECONFIG exists... $EC \n"
   if test -f "$KUBECONFIG"; then
-    echo "$KUBECONFIG exists." && export KUBECONFIG
+    echo "$KUBECONFIG exists, creating $KUBECONFIG_PROXY for proxy use."
+    sed -e '/server:/a\' -e '    proxy-url: socks5://127.0.0.1:1080' $KUBECONFIG > $KUBECONFIG_PROXY
   else
     echo "$KUBECONFIG does not exist." && exit 1
   fi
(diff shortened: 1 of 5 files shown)


fraenkel:
Why not just set a NO_PROXY to access the AWS endpoint?

Secretions:
Strikes me as more correct to not feed awscli a proxy config at all rather than feed it a completely broken one and then `NO_PROXY` the thing we expect it to use. This limits the proxy usage to only the thing that actually needs it (kubectl), and keeps the solution the same as our deployer code (so if something works here, it should work there and vice versa).

steved:
One difficulty for me when crafting a NO_PROXY is that the EKS API URL is very similar to the IAM one so the NO_PROXY has to be incredibly specific.

Secretions:
Deployer probably had more possibilities with what we might run through it, but `sts.amazonaws.com` seems to be enough to get an eks kubeconfig with a role working.

```
[senjougahara : 0 : 3:56PM] terraform>kubectl --kubeconfig=./kubeconfig get po

Failed to connect to proxy URL: "http://socks5://127.0.0.1:1080"
^C
[senjougahara : 130 : 3:56PM] terraform>NO_PROXY=sts.amazonaws.com kubectl --kubeconfig=./kubeconfig get po
No resources found in default namespace.
```

I still think the kubeconfig proxy-url is the more correct solution for the reasons I stated, tho.

fraenkel:
Since we have removed all ENV vars, adding the proxy-url makes sense.
However, why not just use kubectl to do it? 
`kubectl config set-cluster e2e --proxy-url=https://1.2.3.4`

Secretions:
This would look like this:

```
<pass cluster arn to k8s submodule/scripts>
cp kubeconfig kubeconfig-proxied
kubectl --kubeconfig kubeconfig-proxied config set 'clusters.<cluster name which happens to be cluster arn>.proxy-url' https://127.0.0.1:1080
```

Means passing more variables around, and assuming that aws won't change how it happens to choose the cluster name in the kubeconfig (which it probably won't). You can technically also get the cluster name from the kubeconfig itself, but that's actually fiddlier than the sed command.

submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -4,17 +4,20 @@ RED="\e[31m"
...
   ssh -q -N -f -M -o "StrictHostKeyChecking no" -o "ExitOnForwardFailure=yes" -i "${ssh_pvt_key_path}" -D ${k8s_tunnel_port} -S "$TUNNEL_SOCKET_FILE" ${bastion_user}@${bastion_public_ip}
 }
 
 check_kubeconfig() {
-  local KUBECONFIG="${kubeconfig_path}"
   printf "$GREEN Checking if $KUBECONFIG exists... $EC \n"
   if test -f "$KUBECONFIG"; then
-    echo "$KUBECONFIG exists." && export KUBECONFIG
+    echo "$KUBECONFIG exists, creating $KUBECONFIG_PROXY for proxy use."
+    sed -e '/server:/a\' -e '    proxy-url: socks5://127.0.0.1:1080' $KUBECONFIG > $KUBECONFIG_PROXY
```
ddl-dclegg, line 20: Shouldn't the port here (`1080`) actually be a reference to `k8s_tunnel_port`? If not, we should probably make it variable-driven

ddl-dclegg: I guess that goes back to the prior discussion on maintaining or minimizing variable use. 🤔

Secretions: Doh, good call.

Secretions: Github didn't show me your second comment when I responded for some reason, but this is just not using the variables we already have like we should, so I just objectively should have used it as you suggested.

fraenkel (approved):
I still would prefer using kubectl command over sed just to have guarantees across kubectl  upgrades/changes. The sed version makes assumptions about what is currently in the kubeconfig and the order.

Secretions:
There's an assumption in all three solutions that I proposed, but I changed it.

submodules/k8s/variables.tf
```diff
@@ -22,6 +22,11 @@ variable "bastion_public_ip" {
   description = "Bastion host public ip."
 }
 
+variable "eks_cluster_arn" {
```
fraenkel, line 25: we could just ask for the server name.

submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -4,17 +4,21 @@ RED="\e[31m"
...
 }
 
 check_kubeconfig() {
-  local KUBECONFIG="${kubeconfig_path}"
   printf "$GREEN Checking if $KUBECONFIG exists... $EC \n"
   if test -f "$KUBECONFIG"; then
-    echo "$KUBECONFIG exists." && export KUBECONFIG
+    echo "$KUBECONFIG exists, creating $KUBECONFIG_PROXY for proxy use."
+    cp $KUBECONFIG $KUBECONFIG_PROXY
+    kubectl --kubeconfig $KUBECONFIG_PROXY config set "clusters.${eks_cluster_arn}.proxy-url" "socks5://127.0.0.1:${k8s_tunnel_port}"
```
fraenkel, line 21: you could also just use `kubectl set-cluster <cluster name> --proxy-url=....`

submodules/k8s/variables.tf
Secretions: The only ways I could see to get it (without resorting to awscli, which I assume isn't the suggestion) bring us back to text parsing of some sort, since the config options don't have any sort of `jsonpath`-type output options:

```
>kubectl --kubeconfig ./kubeconfig config get-contexts | egrep "^\*" | awk '{ print $3 }'
arn:aws:eks:us-west-2:0123456789:cluster/chibipug6680
>kubectl --kubeconfig ./kubeconfig config get-clusters | egrep -v NAME | head -1
arn:aws:eks:us-west-2:0123456789:cluster/chibipug6680
```

submodules/k8s/templates/k8s-functions.sh.tftpl
Secretions: This only works in kubectls 1.24 and greater, so we'll break 1.23 and earlier provisioning while it's still possible and reduce compatibility on places with a non-current version of kubectl that would have worked if we'd used the more compatible command.

fraenkel: actually we pick the name that we are configuring with.
look at eks/cluster.tf.  We can output the cluster name.

Secretions: I think you might actually be responding to the other comment? I'm going to assume so in my response.

If you mean the `--alias` argument, afaict that only impacts the name of the context. We provide it that argument, and the context gets named the plaintext name of the clsuter, the actual cluster name in the kubeconfig (which we need to reference here) is still the cluster's arn:

```
- cluster:
    certificate-authority-data: <sanitized>
    # proxy-url needs to go here, which is keyed by the arn in `cluster[0].name` below
    server: https://sanitized.gr7.us-west-2.eks.amazonaws.com
  name: arn:aws:eks:us-west-2:1234567890:cluster/chibipug6680
contexts:
- context:
    cluster: arn:aws:eks:us-west-2:1234567890:cluster/chibipug6680
    user: arn:aws:eks:us-west-2:1234567890:cluster/chibipug6680
  name: chibipug6680
current-context: chibipug6680
```

If you mean the `aws_cluster.this.name` attribute, then it's just the incorrect attribute for what we need to target in the kubeconfig

fraenkel: no, just need the cluster arn to be returned. I guess you could return the entire cluster and then just pull the arn as well.

fraenkel: never mind... its what you are doing.  the docs for update-kubeconfig use incorrect wording but in the end it doesn't matter since we are doing the right thing.

PR #26, opened 2023-01-25
PLAT-6183: Setup containerd certs directory on GPU nodes

Secretions:
Although the regular EKS ami sets up containerd with some cert directories, the GPU EKS ami does not.

This either adds a part before the default user data to set that up when no AMI was specified, or adds the same code to our custom user data for custom AMIs.


fraenkel (approved):
We should open an issue against AWS since their AL2 GPU image should just work.

Secretions:
Yeah, I ended up making this a little bit ago: https://github.com/awslabs/amazon-eks-ami/issues/1154

PR #27, opened 2023-02-02
support ECR repositories and set a global KMS key default

steved:
* We'd like to gradually move to ECR for customer installs instead of an internal registry. I did not make it optional.
* Using our own KMS key is [more secure](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-mgmt) than using Amazon-owned keys as they are deployment specific. I kept the BYO KMS key support, but also added the option to use no KMS keys (incl. for EKS).


steved:
Still needs some testing, but good enough for a review.

main.tf
```diff
@@ -136,6 +140,7 @@ module "bastion" {
   public_subnet_id = local.public_subnets[0].subnet_id
   ami_id           = var.bastion.ami
   instance_type    = var.bastion.instance_type
+  kms_key          = var.use_kms ? data.aws_kms_key.key[0].arn : null
```
fraenkel, line 143: why isn't this used above?

main.tf
```diff
@@ -152,13 +157,14 @@ module "eks" {
...
   additional_node_groups       = var.additional_node_groups
-  node_iam_policies            = [module.storage.s3_policy]
+  node_iam_policies            = module.storage.iam_policies
   efs_security_group           = module.storage.efs_security_group
   update_kubeconfig_extra_args = var.update_kubeconfig_extra_args
   eks_master_role_names        = var.eks_master_role_names
   ssh_pvt_key_path             = local.ssh_pvt_key_path
   bastion_user                 = local.bastion_user
   bastion_public_ip            = try(module.bastion[0].public_ip, "")
+  kms_key                      = var.use_kms ? data.aws_kms_key.key[0].arn : null
```
fraenkel, line 167: local.kms_key?

outputs.tf
```diff
@@ -37,3 +37,13 @@ output "kubeconfig" {
   description = "location of kubeconfig"
   value       = local.kubeconfig_path
 }
+
+output "kms_key_id" {
+  description = "KMS key ARN, if enabled"
+  value       = var.use_kms ? data.aws_kms_key.key[0].arn : null
```
fraenkel, line 43: local?

outputs.tf
```diff
@@ -37,3 +37,13 @@ output "kubeconfig" {
   description = "location of kubeconfig"
   value       = local.kubeconfig_path
 }
+
+output "kms_key_id" {
```
fraenkel, line 41: do we always say id when its an arn?
and I see we use key in the var. just need to be consistent so we know what the heck it is.

kms.tf
```diff
@@ -0,0 +1,81 @@
+data "aws_partition" "current" {}
```
fraenkel, line 1: should this be a submodule? can move it later

main.tf
steved: you got hoodwinked by the folding, this was for the bastion. moved it to a local.

outputs.tf
steved: Oddly, yes, and even in external documentation:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/b926e1ee4551e60b44c9b5a887409a8feaf91d86/docs/parameters.md?plain=1#L16
https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#:~:text=key%20using%20the-,x%2Damz%2Dserver%2Dside%2Dencryption%2Daws%2Dkms%2Dkey%2Did,-header.%20If%20the

I was trying to disambiguate with the input since it can be an ID, ARN, alias (whatever the data source can lookup).

fraenkel: well that link shows key-id as the last part of an ARN. 

notice key_id vs key_arn
https://github.com/terraform-aws-modules/terraform-aws-kms/blob/master/outputs.tf#L5

steved: Ok yeah I totally misread it before testing it, I'll add two outputs.

submodules/storage/outputs.tf
```diff
@@ -16,12 +16,17 @@ output "s3_buckets" {
...
 }
 
 output "efs_security_group" {
   description = "EFS security group id"
   value       = aws_security_group.efs.id
 }
+
+output "container_registry" {
+  description = "ECR base registry URL"
+  value       = trimsuffix(aws_ecr_repository.environment.repository_url, "/environment")
```
steved, line 31: Don't know if I love this, but a split, splice, join didn't seem much better

submodules/eks/node-group.tf
```diff
@@ -73,21 +71,12 @@ locals {
...
 }
 
-data "aws_ec2_instance_type" "all" {
-  for_each      = toset(flatten([for ng in local.node_groups : ng.instance_types]))
-  instance_type = each.value
-}
-
-locals {
-  any_gpu = { for name, ng in local.node_groups : name => anytrue([for itype in ng.instance_types : length(data.aws_ec2_instance_type.all[itype].gpus) > 0]) }
-}
```
steved, line 83: Shoving this out of the module is basically the fix for the node group recreation issue, but it means that we don't get autodetection in the module itself

fraenkel: can you elaborate on what exactly triggers it

submodules/storage/outputs.tf
fraenkel: no, I had to track down what you were doing. I get that we are creating 2 private repos with fixed namespaces that map onto what is put together in nucleus. Fragile but it works.

outputs.tf
steved: Nevermind, yeah it's definitely the ARN. The wording is awful, but I think it's the ARN everywhere.

submodules/eks/node-group.tf
steved: Sure, it's mostly described in the ticket; essentially a change to the global tags will trigger a complete recreation of the node groups instead of a minor update to the tags themselves. This seems to be because the `aws_ec2_instance_type` data source is only "known during apply" which means that the `ami_type` for each node group is also deferred until apply and that triggers Terraform's complete recreate policy. I tried a few ways to keep the detection in-module, but ultimately only passing it in triggers the read early enough that it allows the minor updates.

fraenkel: I knew what was triggering it but not the specifics. Getting data resources to resolve during plan is ideal but not always possible. My guess is that the subnet bits deferred everything to apply.

kms.tf
```diff
@@ -76,6 +77,51 @@ resource "aws_kms_alias" "domino" {
...
 
 data "aws_kms_key" "key" {
-  count  = var.use_kms ? 1 : 0
-  key_id = coalesce(var.kms_key_id, try(aws_kms_key.domino[0].arn, null))
+  count  = var.use_kms && var.kms_key_id != null ? 1 : 0
+  key_id = var.kms_key_id
+}
+
+## EKS key
+data "aws_iam_policy_document" "kms_key_eks" {
```
fraenkel, line 85: eks_kms?

what does this have to do with eks?  seems like a generic policy for all kms mgmt.

kms.tf
```diff
@@ -76,6 +77,51 @@ resource "aws_kms_alias" "domino" {
...
+      "kms:List*",
+      "kms:Put*",
+      "kms:Update*",
+      "kms:Revoke*",
+      "kms:Disable*",
+      "kms:Get*",
+      "kms:Delete*",
+      "kms:ScheduleKeyDeletion",
+      "kms:CancelKeyDeletion",
+      "kubeconms:GenerateDataKey",
```
fraenkel, line 102: typo?

submodules/eks/cluster.tf
```diff
@@ -80,7 +39,7 @@ resource "aws_eks_cluster" "this" {
 
   encryption_config {
     provider {
-      key_arn = var.kms_key != null ? var.kms_key : aws_kms_key.eks_cluster.arn
+      key_arn = var.cluster_kms_key
```
fraenkel, line 42: should this be `secrets_kms_key`?

main.tf
```diff
@@ -178,7 +178,8 @@ module "eks" {
   ssh_pvt_key_path             = local.ssh_pvt_key_path
   bastion_user                 = local.bastion_user
   bastion_public_ip            = try(module.bastion[0].public_ip, "")
-  kms_key                      = local.kms_key_arn
+  cluster_kms_key              = var.use_kms ? (var.kms_key_id == null ? resource.aws_kms_key.domino[0].arn : data.aws_kms_key.key[0].arn) : aws_kms_key.eks_cluster[0].arn
```
fraenkel, line 181: this can be simplified since you already computed part of this as the kms_key_arn.
you could also make it more brain dead with a `try(aws_kms_key.eks_cluster[0].arn, local.kms_key_arn)`

fraenkel (approved):

submodules/eks/cluster.tf
```diff
@@ -1,3 +1,44 @@
...
+      "kms:List*",
+      "kms:Put*",
+      "kms:Update*",
+      "kms:Revoke*",
+      "kms:Disable*",
+      "kms:Get*",
+      "kms:Delete*",
+      "kms:ScheduleKeyDeletion",
+      "kms:CancelKeyDeletion",
+      "kubeconms:GenerateDataKey",
```
fraenkel, line 17: typo still

kms.tf
steved: Yeah, I moved this back so it's clearer - for clusters where `use_kms` is false, I don't want to remove the existing KMS encryption for secrets.

outputs.tf
```diff
@@ -40,7 +40,7 @@ output "kubeconfig" {
 
 output "kms_key_id" {
   description = "KMS key ID, if enabled"
-  value       = var.use_kms ? data.aws_kms_key.key[0].id : null
+  value       = var.use_kms ? (var.kms_key_id == null ? resource.aws_kms_key.domino[0].id : data.aws_kms_key.key[0].id) : null
```
fraenkel, line 43: the new model is to use try() instead of checking for nulls

kms.tf
steved: It seems to cause more deferred resolution of the KMS key which can lead to full recreation of some resources (like EFS) during an update to an existing cluster. I don't know if that's a great reason...

fraenkel: must be yet another plan vs apply thing.

submodules/storage/ecr.tf
```diff
@@ -0,0 +1,27 @@
+locals {
+  encryption_type = var.ecr_kms_key != null ? "KMS" : "AES256"
+}
+
```
miguelhar, line 4: These are identical, why not something like 
```hcl

locals {
  ecr_repos = toset(["model","environment"])
}

resource "aws_ecr_repository" "model" {
  for_each             = local.ecr_repos
  name                 = join("/", [var.deploy_id, each.key])
  image_tag_mutability = "IMMUTABLE"

  encryption_configuration {
    encryption_type = local.encryption_type
    kms_key         = var.ecr_kms_key
  }

  force_delete = var.ecr_force_destroy_on_deletion
}


```

submodules/storage/outputs.tf
```diff
@@ -16,12 +16,19 @@ output "s3_buckets" {
   } }
 }
 
-output "s3_policy" {
-  description = "s3 IAM Policy arn"
-  value       = aws_iam_policy.s3.arn
+output "iam_policies" {
+  description = "IAM Policy ARNs"
+  value       = [aws_iam_policy.s3.arn, aws_iam_policy.ecr.arn]
```
miguelhar, line 21: Shouldnt this be a dict so we know which is which?

submodules/storage/variables.tf
```diff
@@ -33,8 +33,26 @@ variable "s3_force_destroy_on_deletion" {
   default     = false
 }
 
-variable "s3_encryption_use_sse_kms_key" {
-  description = "if true use 'aws:kms' else 'AES256' for the s3 server-side-encryption."
+variable "ecr_force_destroy_on_deletion" {
+  description = "Toogle to allow recursive deletion of all objects in the ECR repositories. if 'false' terraform will NOT be able to delete non-empty repositories"
   type        = bool
   default     = false
```
miguelhar, line 39: Should be update our test to make this `true` otherwise we might leak ecr repos

kms.tf
```diff
@@ -0,0 +1,82 @@
...
+
+resource "aws_kms_alias" "domino" {
+  count         = var.use_kms && var.kms_key_id == null ? 1 : 0
+  name          = "alias/${var.deploy_id}"
+  target_key_id = aws_kms_key.domino[0].key_id
+}
+
+data "aws_kms_key" "key" {
+  count  = var.use_kms && var.kms_key_id != null ? 1 : 0
+  key_id = var.kms_key_id
```
Secretions, line 81: We don't do this elsewhere, but is there any reason we couldn't just do the equivalent of `key_id = var.kms_key_id != null ? var.kms_key_id : resource.aws_kms_key.domino[0].id`? I swear I've done that pattern in other terraform before and without it being problematic.

Secretions (approved):

kms.tf
steved: Yeah, I had it like that originally, but it obfuscates the KMS key propagation when passing because the data source can't be read during the plan phase.

submodules/storage/outputs.tf
steved: In this case the policies are blindly attached, do we need to know where they came from?

submodules/storage/variables.tf
steved: I'll create a second PR to set this since the same main.tf.json is used for both sides of the upgrade test

PR #29, opened 2023-02-06
PLAT-6243: Add aws backup vault for efs

Secretions:
Based this on Steve's branch with the KMS changes to avoid moving targets


submodules/storage/variables.tf
```diff
@@ -56,3 +56,39 @@ variable "efs_kms_key" {
   type        = string
   default     = null
 }
+
+variable "efs_backup_vault_kms_key" {
```
fraenkel, line 60: is this the id or arn?

submodules/storage/variables.tf
```diff
@@ -56,3 +56,39 @@ variable "efs_kms_key" {
...
+variable "efs_backup_vault_force_destroy" {
+  description = "Toggle to allow automatic destruction of all backups when destroying."
+  type        = bool
+  default     = false
+}
+
+variable "efs_backup_schedule" {
+  type        = string
+  description = "Cron-style schedule for EFS backup vault"
+  default     = "0 12 * * ? *"
```
fraenkel, line 81: state what the default is in the description

fraenkel (approved):

submodules/storage/variables.tf
Secretions: It's the arn. This is keeping the pattern with the other variables in Steve's PR, so I'll leave any discussion on the variable name for there (and make sure it matches here at the time of merging).

submodules/storage/variables.tf
fraenkel: I meant in English since most folks wont understand cron syntax

PR #32, opened 2023-02-08
add tags_all to ASG node pools as well

steved:
The default provider tags aren't being propagated to node pool instances which breaks our internal cost reporting.


fraenkel (approved):

Secretions (approved):

submodules/eks/node-group.tf
```diff
@@ -111,19 +113,19 @@ resource "aws_launch_template" "node_groups" {
     http_tokens                 = "required"
   }
 
-  # add any tagtag_specifications and additional ones are magically created
+  # add any tag_specifications and additional ones are magically created
   tag_specifications {
```
miguelhar, line 117: Nitpick, but it seems this can be a dynamic block for ["instance","volume"]

miguelhar (approved):

PR #33, opened 2023-02-08
allow selection of # of AZs per node pool

steved:
There's a valid topology where we spread the platform across all AZs, but limit the compute/gpu pools to a single AZ in order to better overprovision and enable faster cold-start performance for users. Unclear yet if this is useful.


fraenkel (approved):

Secretions (commented):
Allowing you to pick the specific availability zones for the individual nodegroups is what we did in CDK, and that was more in response to how certain instance types aren't available in certain AZs, so for tiers depending on those they just want NGs in a/c or what not.

steved:
I can add explicit AZ selection, though the use-case for this is: I want 1 AZ and I don't really care where as long as it is valid.

fraenkel:
https://github.com/dominodatalab/terraform-aws-eks/pull/33#issuecomment-1429000939
The hardest part is not to get caught in the apply side of things which is what tended to happen as we pushed some of this logic around.

PR #34, opened 2023-02-08
allow configuration of bastion SSH IP ranges

steved:
We aren't propagating bastion security group rules from the main module so it's difficult to lock down access to a set of IPs. This assumes you only want to secure the SSH port, so it might not be the best solution.


main.tf
```diff
@@ -143,6 +143,17 @@ module "bastion" {
   ami_id           = var.bastion.ami
   instance_type    = var.bastion.instance_type
   kms_key          = local.kms_key_arn
+  security_group_rules = {
+    bastion_inbound_ssh = {
```
fraenkel, line 147: why is this a map? isn't a list sufficient?

fraenkel (approved):

main.tf
ddl-dclegg: Technically it could be, but I like that a map allows for the key to be descriptive of what the rule does.

submodules/bastion/main.tf
```diff
@@ -22,6 +22,17 @@ resource "aws_security_group" "bastion" {
   }
 }
 
+resource "aws_security_group_rule" "bastion_outbound" {
```
miguelhar, line 25: Should we also make the egress configurable?

submodules/bastion/variables.tf
```diff
@@ -36,7 +36,6 @@ variable "public_subnet_id" {
 }
 
 variable "security_group_rules" {
-
   description = "Bastion host security group rules."
   type = map(object({
```
miguelhar, line 40: Isnt the object missing keys
```hcl
      protocol                 = "tcp"
      from_port                = "22"
      to_port                  = "22"
      type                     = "ingress"
      description              = "Inbound ssh"
      cidr_blocks              = var.bastion.authorized_ssh_ip_ranges
      source_security_group_id = null
```

Secretions (approved):

submodules/bastion/main.tf
steved: I'm not sure, do you think it's ever going to be changed? In the past I've never heard of us needing to change it for security or other reasons

submodules/bastion/variables.tf
steved: Not sure I follow, the type is defined with all those fields. Right now the default is passed in through the main.tf module, though maybe it should be in here?

miguelhar (approved):

PR #35, opened 2023-02-14
allow selection of specific AZs in node pools

steved:
Extension of https://github.com/dominodatalab/terraform-aws-eks/pull/33 to allow specific selection of AZs during node pool creation.


ddl-dclegg (approved):

fraenkel (approved):
The mix and match is kinda weird but at least we are verifying it.

The biggest issue I have with our selecting AZs via specifying count is that all deployments use the same AZs. This causes 2 major issues.
1. We consume the same instance types in the same AZs thereby limiting our total capacity. It may not be a real issue for us, but it can be for customers.
2. An outage will impact all deployments rather than a limited subset.
The issue is that we have no good way to randomize the set of AZs. The random functions are apply only and we know how that ends up. The best way would be to do it outside of TF and provide the list of AZs we wish to use. This should be how our cloud offering does it.

If network peering becomes a thing, one needs to match when using PrivateLink although its AZ ids and not AZ names. For VPC Peering and TGW, you could imagine if the zones are different between the networks, an outage in one would not translate into an outage in the other but its still potentially unusable so you have widened the potential blast radius.
Anyhow its all something to think about.

PR #37, opened 2023-02-14
PLAT-5668: Minor conversion process tweaks

Secretions:
* Don't recreate node security group just because of name/description
* Allow custom aws-auth map role entries
  * Allows us to use original nodegroup iam role so we don't break OG ASGs prior to switchover
  * Mainly for peace of mind (Domino isn't suddenly down)

### DIFF
--- a/submodules/eks/node-group.tf
+++ b/submodules/eks/node-group.tf
@@ -22,6 +22,10 @@ resource "aws_security_group" "eks_nodes" {
 
   lifecycle {
     create_before_destroy = true
+    ignore_changes = [
+        name,
+        description
+    ]
   }
   tags = {
     "Name"                                            = "${local.eks_cluster_name}-eks-nodes"
--- a/submodules/eks/node-group.tf
+++ b/submodules/eks/node-group.tf
@@ -23,8 +23,8 @@ resource "aws_security_group" "eks_nodes" {
   lifecycle {
     create_before_destroy = true
     ignore_changes = [
-        name,
-        description
+      name,
+      description
     ]
   }
   tags = {
--- a/submodules/eks/variables.tf
+++ b/submodules/eks/variables.tf
@@ -144,3 +144,8 @@ variable "node_groups_kms_key" {
   description = "if set, use specified key for the EKS node groups"
   default     = null
 }
+
+variable "eks_custom_role_maps" {
+  type        = list(object({rolearn = string, username = string, groups = list(string)}))
+  description = "Custom role maps for aws auth configmap"
+}
--- a/submodules/eks/variables.tf
+++ b/submodules/eks/variables.tf
@@ -146,6 +146,6 @@ variable "node_groups_kms_key" {
 }
 
 variable "eks_custom_role_maps" {
-  type        = list(object({rolearn = string, username = string, groups = list(string)}))
+  type        = list(object({ rolearn = string, username = string, groups = list(string) }))
   description = "Custom role maps for aws auth configmap"
 }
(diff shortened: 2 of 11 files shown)


submodules/eks/variables.tf
```diff
@@ -144,3 +144,8 @@ variable "node_groups_kms_key" {
   description = "if set, use specified key for the EKS node groups"
   default     = null
 }
+
+variable "eks_custom_role_maps" {
+  type        = list(object({ rolearn = string, username = string, groups = list(string) }))
+  description = "Custom role maps for aws auth configmap"
```
ddl-dclegg, line 150: @Secretions Since this is a "custom..." item, I think we should add a `null` default. If there's a more sensible default, I'm fine with that too (since I see there's no `if` in `submodules/k8s/templates/aws-auth.yaml.tftpl`).

submodules/k8s/variables.tf
```diff
@@ -59,3 +59,8 @@ variable "pod_subnets" {
   type        = list(object({ subnet_id = string, az = string }))
   description = "Pod subnets and az to setup with vpc-cni"
 }
+
+variable "eks_custom_role_maps" {
+  type        = list(object({ rolearn = string, username = string, groups = list(string) }))
+  description = "Custom role maps for aws auth configmap"
```
ddl-dclegg, line 65: We should add a default here too

ddl-dclegg (changes requested):

submodules/eks/variables.tf
Secretions: I copied the default from the root folder's variables to here, which is `[]`. Since it loops through, it seems like the sensible default here. I'm honestly not sure why the default needs to be in all three places, though. Since both the `eks` and `k8s` submodules aren't intended to be used separately, they should always get the input from the root `/variables.tf` and inherit its default. On the other hand, if we change a default, we now have to remember to change it in three places (not that I anticipate ever changing it once this merges).

submodules/eks/node-group.tf
```diff
@@ -22,6 +22,10 @@ resource "aws_security_group" "eks_nodes" {
 
   lifecycle {
     create_before_destroy = true
+    ignore_changes = [
+      name,
```
fraenkel, line 26: why are we ignoring name changes?  that seems very suspect.

fraenkel (approved):

submodules/eks/node-group.tf
Secretions: This is to cause less anxiety in the cdk->tf migration process. If we import the pre-existing node security group, then the old nodes can continue to function and Domino isn't broken at any phase of the process / you can validate that terraform didn't do anything destructive after the initial terraform run.

However, both the name and description fields are immutable and trigger a replacement:

```
  # module.domino_eks.module.eks.aws_security_group.eks_nodes must be replaced
+/- resource "aws_security_group" "eks_nodes" {
      ~ arn                    = "arn:aws:ec2:us-west-2:890728157128:security-group/sg-082c110d78d0a7170" -> (known after apply)
      ~ description            = "jhayes7916/EksStack/UnmanagedSG" -> "EKS cluster Nodes security group" # forces replacement
[...]
      ~ name                   = "jhayes7916-sharedNodeSG" -> "jhayes7916-nodes" # forces replacement
[...]
}
```

Since they are immutable, there should be no circumstance where this is triggered in a security group created by this module, only when importing into state as in the migration process. And since these fields are cosmetic, there should be no harm, either.

steved (approved):

PR #38, opened 2023-02-17
PLAT-6276 track eks version on node_groups

miguelhar:
We were not tracking the k8s version in the node groups. As a result upon upgrade only the eks control plane would upgrade.
[PLAT-6276](https://dominodatalab.atlassian.net/browse/PLAT-6276)


submodules/eks/node-group.tf
```diff
@@ -131,10 +131,20 @@ resource "aws_launch_template" "node_groups" {
...
+data "aws_ssm_parameter" "eks_gpu_ami_release_version" {
+  name = "/aws/service/eks/optimized-ami/${aws_eks_cluster.this.version}/amazon-linux-2-gpu/recommended/release_version"
+}
+
 resource "aws_eks_node_group" "node_groups" {
   depends_on      = [module.k8s_setup]
   for_each        = local.node_groups_by_name
   cluster_name    = aws_eks_cluster.this.name
+  version         = aws_eks_cluster.this.version
+  release_version = startswith(each.key, "gpu") ? nonsensitive(data.aws_ssm_parameter.eks_gpu_ami_release_version.value) : nonsensitive(data.aws_ssm_parameter.eks_ami_release_version.value)
```
steved, line 147: I think this should be `each.value.node_group.gpu`?

fraenkel (approved):
Do we need to think about 
[force_update_version](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#force_update_version)

Secretions (approved):

PR #39, opened 2023-02-21
force selection of AZs IDs for each node pool

steved:
Given our conversation:
* Require AZ selection on each node pool
* I've taken the liberty of switching the convention from name to ID
* There's a bit of a helper on the top-level module, you can specify AZ IDs by index instead of the full ID (`usw2-az[0-9]`)
* Although it's deferred until after the cluster is created, still verify that the given AZs have availability for the instance types:
```
│ Error: Resource precondition failed
│
│   on ../submodules/eks/node-group.tf line 155, in resource "aws_launch_template" "node_groups":
│  155:       condition     = length(setsubtract(each.value.availability_zone_ids, toset(data.aws_ec2_instance_type_offerings.nodes[each.key].locations))) == 0
│     ├────────────────
│     │ data.aws_ec2_instance_type_offerings.nodes is object with 4 attributes
│     │ each.key is "specific_azs"
│     │ each.value.availability_zone_ids is list of string with 1 element
│
│ Instance type(s) ["p3.2xlarge"] for node group "specific_azs" are not available in all the given zones:
│ given = ["usw2-az4"]
│ available = ["usw2-az1","usw2-az2","usw2-az3"]
```


main.tf
```diff
@@ -167,7 +144,8 @@ locals {
     for name, ng in
     merge(var.additional_node_groups, var.default_node_groups) :
     name => merge(ng, {
-      gpu = ng.gpu != null ? ng.gpu : anytrue([for itype in ng.instance_types : length(data.aws_ec2_instance_type.all[itype].gpus) > 0]),
+      gpu                   = ng.gpu != null ? ng.gpu : anytrue([for itype in ng.instance_types : length(data.aws_ec2_instance_type.all[itype].gpus) > 0]),
+      availability_zone_ids = [for az_id in ng.availability_zone_ids : local.zones_by_id[az_id]]
```
steved, line 148: If this doesn't exist the error isn't great:
```
│ Error: Invalid index
│
│   on ../main.tf line 148, in locals:
│  148:       availability_zone_ids = [for az_id in ng.availability_zone_ids : local.zones_by_id[az_id]]
│     ├────────────────
│     │ local.zones_by_id is object with 4 attributes
│
│ The given key does not identify an element in this collection value. An object only supports looking up attributes by name, not by numeric index.
```

main.tf
```diff
@@ -45,15 +23,14 @@ data "aws_subnet" "pod" {
 }
 
 locals {
-  # Get zones where ALL instance types are offered(intersection).
-  zone_intersection_instance_offerings = setintersection([for k, v in data.aws_ec2_instance_type_offerings.nodes : toset(v.locations)]...)
-  # Get the zones that are available and offered in the region for the instance types.
-  az_names    = var.vpc_id != null ? distinct(data.aws_subnet.private[*].availability_zone) : length(var.availability_zones) > 0 ? var.availability_zones : data.aws_availability_zones.available.names
-  offered_azs = setintersection(local.zone_intersection_instance_offerings, toset(local.az_names))
```
steved, line 52: There is the possibility of a mismatch between this old set of randomly selected AZs and the new, sorted set resulting in a recreation of the subnets, like this CircleCI test:

https://app.circleci.com/pipelines/github/dominodatalab/terraform-aws-eks/506/workflows/f9bd7e38-6580-41b0-bcce-0abc05461162/jobs/334

I'm not sure there's any real world usage so I'm not sure if we should bother trying to match up between the two (i.e. sort by name of the AZ instead of ID).

Secretions (approved):

main.tf
```diff
@@ -45,15 +23,14 @@ data "aws_subnet" "pod" {
...
-  num_of_azs  = var.vpc_id != null ? length(local.az_names) : var.number_of_azs
+  zones_by_id = {
```
fraenkel, line 26: why not just zip between names and zone_ids? They are in matching order.

steved: Goal was to provide a short-hand for the top-level module so people don't need to know the region identifier for each AZ. I'm not strongly for this, it's just repetitive.

submodules/eks/variables.tf
```diff
@@ -55,7 +56,7 @@ variable "kubeconfig_path" {
 
 variable "private_subnets" {
   description = "List of Private subnets IDs and AZ"
-  type        = list(object({ subnet_id = string, az = string }))
+  type        = list(object({ subnet_id = string, az = string, az_id = string }))
```
fraenkel, line 59: why do we need the az_id if we have the subnet and we already retrieve the az_id from that?

variables.tf
```diff
@@ -94,13 +69,14 @@ variable "default_node_groups" {
...
-          max_per_az           = optional(number, 10)
-          desired_per_az       = optional(number, 1)
+          ami                   = optional(string, null)
+          bootstrap_extra_args  = optional(string, "")
+          instance_types        = optional(list(string), ["m5.2xlarge"])
+          spot                  = optional(bool, false)
+          min_per_az            = optional(number, 0)
+          max_per_az            = optional(number, 10)
+          desired_per_az        = optional(number, 1)
+          availability_zone_ids = optional(list(number), [1, 2, 3])
```
fraenkel, line 79: this isn't an az zone id...
it should be usw2-az4, no one will understand 1, 2, 3.  Its not what the API returns.

main.tf
steved: Ended up pushing a6d59de for backwards compatibility, it's just easier overall to understand I think.

variables.tf
steved: pushed ff97d6762c554b843e019dcba2ea9b42807c7bd2.

main.tf
steved: Pushed ff97d6762c554b843e019dcba2ea9b42807c7bd2, we can require folks to provide them explicitly.

fraenkel (approved):
looks much cleaner

Secretions (approved):

PR #44, opened 2023-03-01
Trigger setup script on all component changes

Secretions:
Because we don't trigger re-running the setup script on all script/yaml updates, but still do on the root script, in situations where we make changes in the accompanying files we are basically kicking the can down the road to find all the places we broke upgrade compatibility.

In this case, it seems like we broke upgrade compatibility with calico on the recent calico change.


submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -60,7 +60,7 @@ install_calico() {
   echo
 
   printf "$GREEN Installing Calico Custom resources $EC \n"
-  kubectl_cmd create -f - <<EOF
+  kubectl_cmd apply -f - <<EOF
```
miguelhar, line 63: nitpick
```suggestion
  kubectl_apply -f - <<EOF
```

miguelhar (approved):

PR #45, opened 2023-03-02
switch to OIDC for AWS credentials in CI

steved:
Static credentials have a tendency to get stale and OIDC was relatively straightforward to setup. I'll remove the static keys from CircleCI once this is merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,9 +85,6 @@ jobs:
     steps:
       - checkout
       - install_tf
-      - run: |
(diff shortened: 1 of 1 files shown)


fraenkel (commented):
What is the ttl of the creds?

fraenkel (commented):
I would refresh the token between certain steps given its a 1hr token.
There is no guarantee that a create + destroy will happen under 1 hr but we can make some guarantee that each finishes within 1 hour.

fraenkel (approved):

PR #46, opened 2023-03-02
support a configurable bastion username

steved:
The bastion username is currently hardcoded to `ec2-user`.


main.tf
```diff
@@ -161,7 +160,7 @@ module "eks" {
   update_kubeconfig_extra_args = var.update_kubeconfig_extra_args
   eks_master_role_names        = var.eks_master_role_names
   ssh_pvt_key_path             = local.ssh_pvt_key_path
-  bastion_user                 = local.bastion_user
+  bastion_user                 = try(var.bastion.username, null)
```
ddl-dclegg, line 163: Where we're still making this default to `ec2-user` everywhere else, is there any reason we don't do it here?

ddl-dclegg (approved):

main.tf
fraenkel, line 163:
```suggestion
  bastion_user                 = var.bastion.username
```
it always has a value

submodules/k8s/variables.tf
```diff
@@ -9,12 +9,10 @@ variable "ssh_pvt_key_path" {
   description = "SSH private key filepath."
 }
 
-
 variable "bastion_user" {
   type        = string
   description = "ec2 instance user."
   default     = "ec2-user"
```
fraenkel, line 15: nit: it just seems weird that we attempt to set consistent defaults N levels deep rather than make certain things just required.

this bastion username highlights said issue. if we override the username at the top but fail to pass it around, we end up with a broken provision. We might want to think about where defaults truly live.

fraenkel: all other usage should have just been optional(string) except here.

steved: The problem is when the top-level is optional this still passes through `null` and the default on the submodule doesn't kick in.

```
+  bastion_user                 = var.bastion.username
```

fraenkel: ah, the difference between null and not present. so the default here is meaningless.

fraenkel: it just strikes me as odd that we say `string` not optional(string), so a null should either fail or pick the default.  Might be a bug.

steved: TIL about `nullable` - pushed 9819132673ad83dc622b87f93134505e02dad62d

PR #47, opened 2023-03-03
allow configurable public EKS endpoint access

steved:
Allows an optional bastion and public EKS endpoints with mandatory access CIDRs.


submodules/eks/variables.tf
```diff
@@ -151,3 +151,23 @@ variable "eks_custom_role_maps" {
...
+variable "eks_public_access_cidrs" {
+  type        = list(string)
+  description = "EKS API endpoint public access CIDRs"
+  # Default value does not allow any public access
+  default  = ["127.0.0.1/32"]
+  nullable = false
+
+  validation {
+    condition     = length(var.eks_public_access_cidrs) > 0 && !contains(var.eks_public_access_cidrs, "0.0.0.0/0")
+    error_message = "EKS cluster access should not be accessible by the general public"
```
fraenkel, line 171: this seems pretty harsh given I already had to flip it.
But its a pretty weak check. I can easily add 2 CIDRs ( 0.0.0.0/1 + 128.0.0.0/1) that are equivalent to 0.0.0.0/0 so making something a tad harder is questionable.
I would have though that we default to no cidrs whether its [] or null. And when you flip it true, you must encode at least one [ len(var.eks.public_access_cidrs) > 0 ].

miguelhar:
We would need to look into [circle-ip-ranges](https://circleci.com/docs/ip-ranges/?utm_source=google&utm_medium=sem&utm_campaign=sem-google-dg--uscan-en-dsa-tROAS-auth-brand&utm_term=g_-_c__dsa_&utm_content=&gclid=CjwKCAiA3pugBhAwEiwAWFzwdf-nn5tS4rUvvZZAuVcInSIT_avNX95aXPh_RaCvjBcfUb9niem28xoCTRkQAvD_BwE#use-cases) if we plan to use on our ci.

steved:
> We would need to look into [circle-ip-ranges](https://circleci.com/docs/ip-ranges/?utm_source=google&utm_medium=sem&utm_campaign=sem-google-dg--uscan-en-dsa-tROAS-auth-brand&utm_term=g_-_c__dsa_&utm_content=&gclid=CjwKCAiA3pugBhAwEiwAWFzwdf-nn5tS4rUvvZZAuVcInSIT_avNX95aXPh_RaCvjBcfUb9niem28xoCTRkQAvD_BwE#use-cases) if we plan to use on our ci.

Probably not in this repo, but yep, that's the plan :+1:

submodules/eks/variables.tf
steved: Part of the issue is that you can't do validation across multiple variables. I can remove the latter check, just figured it's a decent enough warning.

fraenkel: so one recommendation is to combine the two into one object since they are tied together.

steved: Fair enough, looks pretty clean - 438d569db975ee4403ab7a8164ebad345081d131

fraenkel (approved):

miguelhar (approved):

PR #48, opened 2023-03-03
lookup custom AMI root device name for node groups

steved:
When using a non-EKS AMI, the root device name can be different than `xvda`.


submodules/eks/node-group.tf
```diff
@@ -115,7 +124,7 @@ resource "aws_launch_template" "node_groups" {
   image_id               = each.value.ami
 
   block_device_mappings {
-    device_name = "/dev/xvda"
+    device_name = each.value.ami == null ? "/dev/xvda" : data.aws_ami.custom[each.value.ami].root_device_name
```
fraenkel, line 127: try?

steved: sure

PR #50, opened 2023-03-06
Miguelhar.prov bastion

miguelhar:
We are currently instructing module users to download and install binaries in order to deploy domino. This change aims to reduce possible user error and time by conditionally provisioning the bastion with required binaries.


Secretions (approved):

fraenkel (commented):
how does one update the binaries?
do the binaries get updated if we bump k8s or there are newly available binaries?

miguelhar:
> how does one update the binaries? do the binaries get updated if we bump k8s or there are newly available binaries?

Only `kubectl` its versioned and driven by the k8s version.

PR #51, opened 2023-03-09
Fix username default and typo, bumped eks version

miguelhar:
- Added default for `bastion.username` 
- Fix output typo
- bumped tests eks


outputs.tf
```diff
@@ -1,6 +1,6 @@
 output "ssh_bastion_command" {
```
steved, line 1: Should this be an output of the EKS module? The idea was to keep the default username in one place

submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -8,10 +8,12 @@ KUBECONFIG="${kubeconfig_path}"
 KUBECONFIG_PROXY="$KUBECONFIG-proxy"
 
 open_ssh_tunnel_to_k8s_api() {
-  if [[ -n "${bastion_public_ip}" ]]; then
+  if [[ -n "${bastion_public_ip}" && -n ${bastion_user} ]]; then
```
steved, line 11: should be quoted I think? otherwise it's `[[ -n "1.2.3.4" && -n ]]; then`

main.tf
```diff
@@ -157,7 +157,7 @@ module "eks" {
   create_bastion_sg            = var.bastion != null
   bastion_security_group_id    = try(module.bastion[0].security_group_id, null)
   bastion_public_ip            = try(module.bastion[0].public_ip, null)
-  bastion_user                 = try(var.bastion.username, null)
+  bastion_user                 = try(module.bastion[0].user, null)
   kubeconfig_path              = local.kubeconfig_path
   node_groups                  = local.node_groups
   node_iam_policies            = module.storage.iam_policies
```
steved, line 163: Not the actual place for the comment, but does `module "bastion'` need a reference to `var.bastion.user`?

miguelhar: bastion takes the user 
```hcl
variable "bastion_user" {
  type        = string
  description = "ec2 instance user."
  default     = "ec2-user"

}
```

I think is the bastion module what should determine what the default user should be, so we should just source it

miguelhar: This goes back to my comment where we should be uniform in the structure of the vars.

outputs.tf
miguelhar: I believe it should be an output of the bastion itself, moved there

main.tf
steved: I don't think we're passing the variable from `var.bastion` at the top-level is all:
```
diff --git a/main.tf b/main.tf
index 5291ef3..d72daad 100644
--- a/main.tf
+++ b/main.tf
@@ -113,6 +113,7 @@ module "bastion" {
   instance_type     = var.bastion.instance_type
   kms_key           = local.kms_key_arn
   install_binaries  = var.bastion.install_binaries
+  bastion_user      = var.bastion.username
   k8s_version       = var.k8s_version
   security_group_rules = {
     bastion_inbound_ssh = {
```

submodules/k8s/main.tf
```diff
@@ -23,8 +23,8 @@ locals {
         aws_auth_yaml       = basename(local.aws_auth_filename)
         eniconfig_yaml      = local.eniconfig_filename != "" ? basename(local.eniconfig_filename) : ""
         calico_operator_url = local.calico.operator_url
-        bastion_user        = var.bastion_user
-        bastion_public_ip   = var.bastion_public_ip
+        bastion_user        = var.bastion_user != null ? var.bastion_user : ""
+        bastion_public_ip   = var.bastion_public_ip != null ? var.bastion_public_ip : ""
```
steved, line 27: FWIW `var.bastion_public_ip` has `nullable = false` which why this was fine before

main.tf
steved: I still think this needs to be addressed ^

submodules/k8s/main.tf
```diff
@@ -23,8 +23,8 @@ locals {
         aws_auth_yaml       = basename(local.aws_auth_filename)
         eniconfig_yaml      = local.eniconfig_filename != "" ? basename(local.eniconfig_filename) : ""
         calico_operator_url = local.calico.operator_url
-        bastion_user        = var.bastion_user
-        bastion_public_ip   = var.bastion_public_ip
+        bastion_user        = var.bastion_user != null ? var.bastion_user : ""
```
steved, line 26: `coalesce(var.bastion_user, "")`?

steved (approved):

submodules/k8s/main.tf
```diff
@@ -23,7 +23,7 @@ locals {
         aws_auth_yaml       = basename(local.aws_auth_filename)
         eniconfig_yaml      = local.eniconfig_filename != "" ? basename(local.eniconfig_filename) : ""
         calico_operator_url = local.calico.operator_url
-        bastion_user        = var.bastion_user != null ? var.bastion_user : ""
+        bastion_user        = coalesce(var.bastion_user, "")
```
fraenkel, line 26: what is the answer when var.bastion_user is null or ""? coalesce ignores both null and "".
if this is the answer then why aren't we doing the same for bastion_public_ip?

submodules/k8s/main.tf
miguelhar: coalesce ignores both `null` and `""` 
```
> coalesce(null, "")
╷
│ Error: Error in function call
│
│   on <console-input> line 1:
│   (source code not available)
│
│ Call to function "coalesce" failed: no non-null, non-empty-string arguments.
╵
```
we need the empty string to make it to the template

steved: yep, my bad

PR #57, opened 2023-03-15
update defaults to match previous production defaults

steved:
Our previous defaults differed slightly:
* 1000GB root volumes for compute nodes: https://github.com/dominodatalab/cdk-cf-eks/blob/542a0c361d9d76470c95a391cb085ff586f631f4/cdk/domino_cdk/config/template.py#L30
* m5.2xlarge platform if Istio was disabled: https://github.com/dominodatalab/cdk-cf-eks/blob/master/cdk/domino_cdk/config/template.py#L31
* 0 nodes for compute pools by default

The only one I feel strongly about is the EBS root volume. I think we'd likely waste some money if m5.4xlarge is the default even with Istio - we use 4 nodes with 2xlarge still.


variables.tf
```diff
@@ -75,7 +75,7 @@ variable "default_node_groups" {
           spot                  = optional(bool, false)
           min_per_az            = optional(number, 0)
           max_per_az            = optional(number, 10)
-          desired_per_az        = optional(number, 1)
+          desired_per_az        = optional(number, 0)
```
fraenkel, line 78: this is a bit onerous given we need at least 1 compute node. Now there is an install timing issue that we may spin one up and then have it spin down because min = 0 and it may take 10+ minutes before we install anything on the node.
I guess there is a difference between having 3 nodes (1 per AZ) spin up when we only use 1 of them.

fraenkel (approved):

Secretions (approved):

PR #59, opened 2023-03-20
SSM cloudwatch integration

domino-pablo-rodriguez:
This PR adds some permissions to push the session logs up to Cloudwatch for Panther integration:

https://docs.aws.amazon.com/systems-manager/latest/userguide/getting-started-create-iam-instance-profile.html#create-iam-instance-profile-ssn-logging

Related PR: https://github.com/cerebrotech/stamp/pull/149


steved (approved):
+1 from me as long as any of @fraenkel's concerns are addressed

fraenkel:
I think we have to tread lightly here. We shouldn't increase the IAM permissions unless there is a direct Customer need.
We have two choices, either its guarded behind a flag or there is a separate module that cloud uses to drive all additional requirements. Given there is only one right now, I would opt for the easiest solution but once we add the second, we should reconsider how Cloud should be done given it has additional unrelated concerns.

domino-pablo-rodriguez:
Thanks for the reply @fraenkel. I'm still unsure why do you think there's no customer value. We are already enabling the required IAM permission for SSM to work without a feature flag, but you are requesting a feature flag to enable audit logging the connections for the SSM agent, is this correct? Or are you suggesting that both things live behind a FF?

submodules/eks/iam.tf
```diff
@@ -191,11 +191,35 @@ data "aws_iam_policy_document" "snapshot" {
...
+    effect    = "Allow"
+    resources = ["*"]
+
+    actions = [
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams"
+    ]
+  }
```
steved, line 205: Can we limit these permissions a bit more granularly?

https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-rc-setting-up-cwlogs.html

steved (approved):
IMO this doesn't need extra justification given the scope of the permissions and a bit more granularity I commented on, but I think it's a point well noted.

submodules/eks/variables.tf
```diff
@@ -134,6 +134,12 @@ variable "bastion_public_ip" {
   default     = null
 }
 
+variable "ssm_log_group_name" {
```
steved, line 137: You'll probably need to pass this through the top-level `variables.tf` to make it truly configurable

submodules/eks/iam.tf
steved: Works for me!

steved (approved):

PR #66, opened 2023-04-03
Miguelhar.var fix

miguelhar:
- Numbered examples for concurrent execution.
- `var.region` for provider setup.
- `bastion.enabled` for more direct conditional, reduces errors `The "for_each" map includes keys derived from resource attributes that cannot be determined until apply`.
- random port for k8s tunnel, allows parallel execution.
- Bastion `root_block_device` [kms_key_id](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#kms_key_id) takes the `arn`.
- `ssm_log_group_name` -> `eks` object.
- `kms` enabled bydefault.

[PLAT-6452](https://dominodatalab.atlassian.net/browse/PLAT-6452)


submodules/eks/main.tf
```diff
@@ -153,6 +153,7 @@ locals {
 
   eks_info = {
     cluster = {
+      public_access     = var.eks.public_access
```
ddl-dclegg, line 156: Should we put a default of `{}` here? If I'm not mistaken, `var.eks` can potentially be just `{}`, and this should then throw an error.

miguelhar:
test-upgrade is expected to fail.

fraenkel (approved):

ddl-dclegg (approved):
I just have a few questions, and we probably need to standardize this PR name and put `<ticket> - <descriptive name>`

submodules/eks/main.tf
miguelhar: Defaults are set in the root vars which pass through to the submodule.
https://github.com/dominodatalab/terraform-aws-eks/blob/8d692ba55c7aa980226060a416d8f282c662d1e6/variables.tf#L79
we do not expect an empty `{}` `eks` from the root module

PR #68, opened 2023-04-04
Ci loads vars from terraform.tfvars instead of env var.

miguelhar:
Load tf vars from file instead of env.


ddl-dclegg (approved):
Looks good to me

PR #69, opened 2023-04-05
Adds a simple sh to run a tf plan on the examples

miguelhar:
`examples/tf-plan-test.sh` will be incorporated into CI later as well as testing `bring-your-vpc` .

### DIFF
--- a/examples/tf-plan-test.sh
+++ b/examples/tf-plan-test.sh
@@ -1,0 +1,48 @@
+#! /usr/bin/env bash
+
+exclude=("bring-your-vpc" "examples.pem" "tf-plan-test.sh")
+
+failed_dirs=()
+success_dirs=()
+
+verify_terraform() {
+  if ! [ -x "$(command -v terraform)" ]; then
+    echo "Error: Terraform is not installed."
+    exit 1
+  else
+    terraform_version=$(terraform --version | awk '/Terraform/ {print $2}')
+    echo "Terraform version $terraform_version is installed."
+  fi
+}
+
+run_terraform_plans() {
+  for dir in */; do
+    [[ " ${exclude[*]} " == *" ${dir%/} "* ]] && continue
+    printf "\n\033[0;33mRunning terraform plan in $dir\033[0m:\n"
+
+    terraform -chdir="$dir" init
+    terraform -chdir="$dir" plan
+
+    if [ "$?" != "0" ]; then
+      printf "\033[0;31mERROR: terraform plan failed in $dir\033[0m.\n"
+      failed_dirs+=("$dir")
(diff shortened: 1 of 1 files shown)


examples/tf-plan-test.sh
```diff
@@ -0,0 +1,48 @@
...
+      failed_dirs+=("$dir")
+    else
+      printf "\033[0;32mSUCCESS: terraform plan succeeded in $dir\033[0m.\n"
+      success_dirs+=("$dir")
+    fi
+  done
+
+  if [ "${#failed_dirs[@]}" != "0" ]; then
+    printf "\n\033[0;31mThe following examples failed the terraform plan:\033[0m\n"
+    printf '\033[0;31m%s\n\033[0m' "${failed_dirs[@]} "
```
ddl-dclegg, line 37: @miguelhar This will only give the first value, I believe you mean `"${failed_dirs[*]}`

examples/tf-plan-test.sh
```diff
@@ -0,0 +1,48 @@
...
+    exit 1
+  fi
+
+  if [ "${#success_dirs[@]}" != "0" ]; then
+    printf "\n\033[0;32mThe following examples ran the terraform plan successfully:\033[0m\n"
+    printf '\033[0;32m%s\n\033[0m' "${success_dirs[@]}"
```
ddl-dclegg, line 43: Should probably be `"${success_dirs[*]}"` to list all

ddl-dclegg (changes requested):

ddl-dclegg:
For the record, other than my 2 comments, this is 👍 from me

examples/tf-plan-test.sh
miguelhar: i believe `${failed_dirs[*]}` is used when elements contain white space or special chars which will be used to split

ddl-dclegg (approved):
If the tests show the desired behavior, that's fine by me. When I ran shellcheck on this script, seeing this error
image
I checked into things and had the following results with `echo` vs. `printf`
```bash
$ ARRTEST=( "/" "/tmp" "/usr/local" )
$ echo "${ARRTEST[@]}"
/ /tmp /usr/local
$ echo "${ARRTEST[*]}"
/ /tmp /usr/local

$ printf "${ARRTEST[@]}"
/
$ printf "${ARRTEST[*]}"
/ /tmp /usr/local
```

examples/tf-plan-test.sh
ddl-dclegg: Please see my latest general comment. If you are satisfied with your current results, we can mark these comments "resolved"

PR #73, opened 2023-04-08
Run examples plan test in CI

miguelhar:


.circleci/config.yml
```diff
@@ -98,7 +98,27 @@ commands:
...
+  tf_plan_test:
+    steps:
+      - aws-cli/setup:
+         role-arn: "${AWS_IAM_ROLE}"
+         session-duration: "900"
+      - run:
+          name: Terraform plan test
+          working_directory: examples
+          command: |
+            bash tf-plan-test.sh
```
fraenkel, line 110: if you require bash, why isn't it a shebang in the tf-plan-test.sh?

.circleci/config.yml
```diff
@@ -98,7 +98,27 @@ commands:
...
 jobs:
+  tf-plan-test:
+    docker:
+      - image: cimg/aws:2022.11.1
```
fraenkel, line 114: why arent we using `2023.04`?

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ parameters:
     default: "v3.10.0"
     type: string
   terraform_version:
-    default: "1.3.7"
+    default: "1.4.2"
```
fraenkel, line 8: why not go to 1.4.4?

.circleci/config.yml
```diff
@@ -98,7 +98,27 @@ commands:
...
 jobs:
+  tf-plan-test:
+    docker:
+      - image: cimg/aws:2022.11.1
+    parameters:
+      terraform_version:
+        type: string
+    steps:
+      - checkout
+      - install_tf
```
fraenkel, line 120: don't you want to pass the terraform_version?

fraenkel (approved):

.circleci/config.yml
miguelhar: It is https://github.com/dominodatalab/terraform-aws-eks/blob/5896cf2bd812b758d921f004776d564e55df1251/examples/tf-plan-test.sh#L1.

fraenkel: so why bash it?

.circleci/config.yml
miguelhar: yes.

.circleci/config.yml
miguelhar: I do
https://github.com/dominodatalab/terraform-aws-eks/blob/5896cf2bd812b758d921f004776d564e55df1251/.circleci/config.yml#L213
https://github.com/dominodatalab/terraform-aws-eks/blob/5896cf2bd812b758d921f004776d564e55df1251/.circleci/config.yml#L116

fraenkel: you aren't passing it, you are just using defaults everywhere.

fraenkel (approved):

PR #77, opened 2023-04-19
[DOM-46093] Adding support for prefix for template files to avoid file generation collision

ldebello-ddl:
# Issue

The submodule "eks" is referencing the same submodule twice which leads to a templates files being override, we need to keep both files because those are stored in s3 and we want to re-run tf for infra upgrades and os patching.

# Proposal

Include a prefix to the template being generated

**Open to other approaches in case this is not optimal.**


miguelhar:
The current issue is that the files are generated twice, but they are exactly the same, given that execution is serial there should not be a race condition. Rather than isolating the generated files, we should provide a solution that only executes the templates once.

miguelhar (changes requested):
we can create a bool `render_templates` which we set on the initial call

https://github.com/dominodatalab/terraform-aws-eks/blob/main/submodules/k8s/main.tf#L66
```
or_each             = { for k, v in local.templates : k => v if v.filename != "" && var.render_templates }
```

ldebello-ddl:
> The current issue is that the files are generated twice, but they are exactly the same, given that execution is serial there should not be a race condition. Rather than isolating the generated files, we should provide a solution that only executes the templates once.

Yes and no, as of today 3 out of 4 files render the same output (excepting "k8s-pre-setup.sh" because the flags "cluster_setup" and "install_calico" are configured with different values), but giving the fact the 4 files are templates the same issue could happen to any of them.

Secretions:
@ldebello-ddl @miguelhar : As the guy who made this mess, I had some thoughts on how to fix it and have an alternate suggestion in this PR: https://github.com/dominodatalab/terraform-aws-eks/pull/79

ldebello-ddl:
> @ldebello-ddl @miguelhar : As the guy who made this mess, I had some thoughts on how to fix it and have an alternate suggestion in this PR: #79

Thanks, I will tested it

ldebello-ddl:
closing this PR in favor of https://github.com/dominodatalab/terraform-aws-eks/pull/79, thanks @Secretions for the help

PR #79, opened 2023-04-21
Refactor split submodule scripts

Secretions:
To push Calico install to the end in #71, I referenced the k8s submodule twice. This was intentional, even though it was a bit messy, as we were simply rewriting the same template files, but this apparently is causing issues with s3 as a store.

This takes a hybrid approach, keeping the modularization of the scripts overall, but putting the actual execution into the EKS module.

### DIFF
--- a/submodules/eks/k8s.tf
+++ b/submodules/eks/k8s.tf
@@ -14,12 +14,12 @@ module "k8s_setup" {
   depends_on = [aws_eks_addon.vpc_cni, null_resource.kubeconfig]
 }
 
-resource "null_resource" "run_k8s_pre_setup" {
+resource "terraform_data" "run_k8s_pre_setup" {
   count = local.run_setup
 
-  triggers = {
-    change_hash = module.k8s_setup[0].change_hash
-  }
+  triggers_replace = [
+    module.k8s_setup[0].change_hash
+  ]
 
   provisioner "local-exec" {
     command     = "./${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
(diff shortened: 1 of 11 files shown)


submodules/eks/k8s.tf
```diff
@@ -1,25 +1,47 @@
...
+  bastion_info = var.bastion_info
+  network_info = var.network_info
+  eks_info     = local.eks_info
 
   depends_on = [aws_eks_addon.vpc_cni, null_resource.kubeconfig]
 }
 
-module "calico_setup" {
-  count = var.bastion_info != null || var.eks.public_access.enabled ? 1 : 0
+resource "null_resource" "run_k8s_pre_setup" {
```
miguelhar, line 17: Can we run this one as part of the `k8s_setup` otherwise `k8s_setup` is only rendering the templates

miguelhar (changes requested):

submodules/k8s/main.tf
```diff
@@ -70,21 +68,6 @@ resource "local_file" "templates" {
...
-    command     = basename(local_file.templates["k8s_presetup"].filename)
-    interpreter = ["bash"]
-    working_dir = local.resources_directory
-  }
-
-  depends_on = [
-    local_file.templates,
-  ]
+locals {
+  change_hash = "${md5(local_file.templates["k8s_presetup"].content)}-${md5(local_file.templates["k8s_functions_sh"].content)}-${md5(local_file.templates["aws_auth"].content)}-${try(md5(local_file.templates["eni_config"].content), "none")}"
```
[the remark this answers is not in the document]
Secretions: I did something halfway in between so there's no risk of four `none`s.

```
locals {
  change_hash = "${join("-", [for file in ["k8s_presetup", "k8s_functions_sh", "aws_auth"] : md5(local_file.templates[file].content)])}-${try(md5(local_file.templates["eni_config"].content), "none")}"
}
```

submodules/eks/k8s.tf
Secretions: I had it like that after the initial change, but it seemed like kind of a strange division of labor to have the script running there, and then arbitrarily (if still for a reason) running it again outside of the module, including all the bits to run it correctly externally.

ldebello-ddl (approved):
LGTM

ldebello-ddl:
@Secretions @miguelhar could we merge this PR?

miguelhar (approved):

PR #81, opened 2023-04-21
Miguelhar.plat 6528 OIDC support

miguelhar:
[PLAT-6528](https://dominodatalab.atlassian.net/browse/PLAT-6528)


submodules/storage/iam.tf
```diff
@@ -14,7 +14,36 @@ data "aws_iam_policy_document" "s3" {
...
+
+    actions = [
+      "s3:ListBucket",
+      "s3:GetBucketLocation",
+      "s3:ListBucketMultipartUploads",
+    ]
+  }
+
+  statement {
+    sid    = ""
```
ddl-dclegg, line 43: Nit: why not actually give it a statement id instead of an empty string?

submodules/storage/variables.tf
```diff
@@ -116,3 +116,11 @@ variable "network_info" {
     })
   })
 }
+
+variable "irsa" {
+  description = "IRSA object."
```
ddl-dclegg, line 121: Can we make this more descriptive? I realize someone setting up IRSA already knows, but it would be good to save people who don't a search for definitions or docs.

submodules/irsa/variables.tf
```diff
@@ -0,0 +1,85 @@
...
+  description = "Domino Deployment ID"
+
+  validation {
+    condition     = can(regex("^[a-z-0-9]{3,32}$", var.deploy_id))
+    error_message = "Argument deploy_id must: start with a letter, contain lowercase alphanumeric characters(can contain hyphens[-]) with length between 3 and 32 characters."
+  }
+}
+
+variable "storage_info" {
+  description = "Storage info."
```
ddl-dclegg, line 12: This is pretty generic, we should add a little more declarative information to clarify its purpose.

ddl-dclegg (changes requested):
Looks good generally, we just need to pass the linting and clarify a few things.

PR #84, opened 2023-04-28
PLAT-6619: Fix user-provided KMS & don't provision unused EKS KMS key

Secretions:
* User-provided KMS keys being set to the id when everything downstream consumes the actual object, breaking user-provided kms support
* We always provision an EKS-specific KMS key, even though we only ever use it in the circumstance where KMS is disabled
  * The most straightforward solution would be a toggle on that
  * But since we already provision a KMS key as a matter of course, I instead removed the EKS-only one and scoped it only to the EKS cluster when KMS is disabled
* IMO we should start testing with a provided KMS key. The non-provided case will get plenty of use elsewhere (and likely get caught during development).
  * This was secretly broken in CDK, and now was secretly broken here.
  * However, I don't want to introduce the test in this PR because the upgrade test would fail on the previous version not having it. I will add the test in a subsequent PR.


kms.tf
```diff
@@ -6,7 +6,7 @@ locals {
 }
 
 data "aws_iam_policy_document" "kms_key_global" {
-  count = var.kms.enabled && var.kms.key_id == null ? 1 : 0
+  count = local.create_kms_key
```
Secretions, line 9: @miguelhar Was this not using the local variable for any particular reason? Want to make sure I'm not overlooking some circumstance where terraform will evaluate this as unable to know until apply or something.

Secretions: Didn't initially run into it when testing, but trying to integrate tests in another branch and actually hitting it. The localness of the variable doesn't make a difference here, but I may need to change this to a list of strings or change how my test is made and provision the custom kms key to test with separately (which kind of sucks).

variables.tf
```diff
@@ -339,21 +339,21 @@ variable "storage" {
 variable "kms" {
   description = <<EOF
     enabled = "Toggle,if set use either the specified KMS key_id or a Domino-generated one"
-    key_id  = optional(string, null)
+    key_id  = optional(list(string), [])
```
miguelhar, line 342: why does the key id need a list?

main.tf
```diff
@@ -84,6 +84,7 @@ module "eks" {
   eks                = var.eks
   network_info       = module.network.info
   kms_info           = local.kms_info
+  kms_enabled        = var.kms.enabled
```
miguelhar, line 87: should this be in `kms_info`?

variables.tf
Secretions: Explained it in the comment where I pinged you, but not very clearly. To be able to do tests, I set up the test to provision a KMS key outside of the module. However, because we're using the key with a `count` and the key's value cannot be known to terraform prior to apply it either errors or requires a targeted apply of the before running "for real". By making it a list of strings, we can check the length of the list, getting around this terraform restriction.

While I don't like doing obtuse things just for the sake of the test, it also occurs to me that perhaps this change is positive since someone consuming this terraform could conceivably want to do something similar.

If we don't want to do this, I can also just switch the test to do a targeted apply or configure the external kms key with separate terraform (or if you know a better way around this, we can do that).

main.tf
Secretions: Yeah, baked it in.

PR #85, opened 2023-04-28
PLAT-6619: Tests for KMS keys

Secretions:
This is a PR to add customer-provisioned KMS to the CI tests, as followup to #84. It's separate since the upgrade test would be upgrading from a version of this module that's broken and fail, so this way we can create a release with the fixed version and have the upgrade test succeed "for real" prior to merge.


.circleci/config.yml
```diff
@@ -176,7 +176,8 @@ jobs:
           name: "Set module source to latest published release"
           working_directory: tests
           command: |
-            latest_release_tag="$(git tag --sort=-v:refname | head -1)"
+            # latest_release_tag="$(git tag --sort=-v:refname | head -1)"
+            export latest_release_tag="PLAT-6503-ish"
```
Secretions, line 180: This will be reverted once #84 is merged and a release is created based on it, prior to merging this.

ddl-dclegg (approved):

PR #86, opened 2023-05-01
Catch empty lookup for bastion_info

miguelhar:


main.tf
```diff
@@ -6,7 +6,7 @@ locals {
     key_id  = local.kms_key.id
     key_arn = local.kms_key.arn
   } : null
-  bastion_info = var.bastion.enabled ? module.bastion[0].info : null
+  bastion_info = var.bastion.enabled ? try(module.bastion[0].info, null) : null
```
fraenkel, line 9: isnt that redundant.  the default value on a try is null
in theory the ternary isn't needed either since when not enabled it should fail and be null

fraenkel (approved):

main.tf
miguelhar: the problem we run is that we are looking at 2 different places `var.bastion.enabled` and `module.bastion[0].info` given that `var.bastion.enabled` can always be evaluated, such is not the case for `module.bastion[0].info` and in the absence of the try this fails with:
```
2023-04-28 15:21:29,679 -     │ module.bastion is empty tuple
```
the purpose of the try is to catch this without blowing up.

miguelhar: this will not work as we break the lifecycle 

```
Plan: 148 to add, 0 to change, 0 to destroy.
╷
│ Error: Invalid for_each argument
│ 
│   on ../submodules/eks/eks-bastion-access.tf line 43, in resource "aws_security_group_rule" "bastion_eks":
│   43:   for_each = local.bastion_eks_security_group_rules
│     ├────────────────
│     │ local.bastion_eks_security_group_rules is a map of object, known only after apply
│ 
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify the instances of this resource.
│ 
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place apply-time results only in the map values.
│ 
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and then apply a second time to fully converge.
╵
╷
│ Error: Invalid count argument
│ 
│   on ../submodules/eks/k8s.tf line 6, in module "k8s_setup":
│    6:   count = local.run_setup
│ 
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the count depends on.
```

PR #87, opened 2023-05-01
PLAT-6614 handle kms.enabled

miguelhar:
[PLAT-6614](https://dominodatalab.atlassian.net/browse/PLAT-6614)

Ignore kms changes to avoid resources destruction.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,7 +60,7 @@ repos:
         args:
           - '--args=--compact'
           - '--args=--quiet'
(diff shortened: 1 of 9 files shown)


fraenkel (approved):

examples/kms/main.tf
```diff
@@ -19,6 +19,8 @@ module "domino_eks" {
     enabled = true
   }
   kms = {
-    enabled = true
+    enabled = false
```
Secretions, line 22: Couldn't we just lifecycle ignore this so we can have the default we want? Seems really unfortunate to debut with a default other than our preference due to pre-release usage.

Secretions (changes requested):

examples/kms/main.tf
miguelhar: Given that this is a very intrusive setting I rather we are explicit when enabling it. But I do agree having the lifecycle ignore its a good additional settings given that we are not looking to change the behavior once set

Secretions: Ok, if we legitimately want the default to be off that's fine. I thought we were sacrificing a prior decision about defaults for compatibility.

Secretions (approved):
Long as we proactively want kms to be default, I'm fine with this.

ddl-dclegg (approved):
I'm good with this route forward.

Secretions (approved):

PR #92, opened 2023-05-16
Miguelhar.plat 6528 OIDC

miguelhar:
[PLAT-6528](https://dominodatalab.atlassian.net/browse/PLAT-6528)


submodules/eks/cluster.tf
```diff
@@ -72,6 +72,17 @@ resource "aws_eks_cluster" "this" {
   }
 }
 
+data "tls_certificate" "cluster_tls_certificate" {
+  url = aws_eks_cluster.this.identity[0].oidc[0].issuer
```
ddl-dclegg, line 76: Can we do something like `try(aws_eks_cluster.this.identity[0].oidc[0].issuer, "")` here? In some recent operations I've seen index errors on things like this when the reference/attribute is null.

fraenkel:
This just seems to be the first part of the actual task. You have enabled OIDC to EKS.
How does one configure the mapping?

miguelhar:
Thats we discussed we were doing. Thats up to the consumer to set up for the time being. While we figure out the structure.

submodules/eks/cluster.tf
miguelhar: Thats a property on eks I dont expect it to ever change and if it does I rather it blows up.

steved (approved):
One question, but LGTM as a first step.

submodules/eks/cluster.tf
ddl-dclegg: I'm not sure I agree, but I'm not going to hold things up because of that.

ddl-dclegg (approved):

PR #94, opened 2023-05-22
PLAT-6617 Use a permanent role to create eks.

miguelhar:
[PLAT-6617](https://dominodatalab.atlassian.net/browse/PLAT-6617)


iam.tf
```diff
@@ -37,3 +37,69 @@ resource "aws_iam_role_policy_attachment" "route53" {
   policy_arn = aws_iam_policy.route53[0].arn
   role       = lookup(module.eks.info.nodes.roles[count.index], "name")
 }
+
+data "aws_iam_policy_document" "create_eks_role" {
+  statement {
+    sid = "EKSDeployerEKS"
+    actions = [
+      "eks:*"
```
miguelhar, line 45: Given that we are close to `*` https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonelastickubernetesservice.html .

iam.tf
```diff
@@ -37,3 +37,77 @@ resource "aws_iam_role_policy_attachment" "route53" {
...
+    actions = [
+      "iam:PassRole",
+      "iam:GetRole"
+    ]
+    resources = ["arn:${data.aws_partition.current.partition}:iam::${local.aws_account_id}:role/${var.deploy_id}-*"]
+    effect    = "Allow"
+  }
+}
+
+resource "aws_iam_policy" "create_eks_role" {
```
fraenkel, line 80: this is gonna be a problem in the same account.
how do we deploy 2 deployments?

fraenkel (changes requested):

iam.tf
Secretions: I think the `local.create_eks_role_name` below should cover that?

fraenkel: yes I missed the name was a local var.

tests/main.tf.json
```diff
@@ -80,7 +80,7 @@
       "deploy_id": {
         "type": "string",
         "description": "Domino Deployment ID.",
-        "default": "dominotfeks"
+        "default": "dominotfeks2"
```
fraenkel, line 83: why bother with a default?

versions.tf
```diff
@@ -22,3 +26,18 @@ provider "aws" {
...
+  alias  = "eks"
+  region = var.region
+  default_tags {
+    tags = var.tags
+  }
+
+  assume_role {
+    # https://github.com/hashicorp/terraform/issues/30690
+    # https://github.com/hashicorp/terraform/issues/2430
+    role_arn = "${aws_iam_role.create_eks_role.arn}${time_sleep.create_eks_role_10_seconds.id == "nil" ? "" : ""}"
```
fraenkel, line 41: this is just another highlight that we should be building this as separate TF steps.

fraenkel (approved):

Secretions (approved):
Should be worth noting that anyone who adopts in 5.6.0 (or is otherwise upgrading from before this commit) is going to need to add this to their config to survive the kubeconfig getting recreated, should that get triggered:

```
eks = {
    custom_role_maps = [{
            rolearn = "arn:aws:iam::<account-id>:role/<deploy-name>-create-eks"
            username = "<deploy-name>-create-eks"
            groups = [
              "system:masters"
            ]
        }]
}
```

I think that should pretty much just work, but if somehow recerating the kubeconfig was triggered before adding that, they'll need to use a version of the kubeconfig without the role to edit the aws-auth configmap temporarily to add this so they can proceed.

miguelhar:
> Should be worth noting that anyone who adopts in 5.6.0 (or is otherwise upgrading from before this commit) is going to need to add this to their config to survive the kubeconfig getting recreated, should that get triggered:
> 
> ```
> eks = {
>     custom_role_maps = [{
>             rolearn = "arn:aws:iam::<account-id>:role/<deploy-name>-create-eks"
>             username = "<deploy-name>-create-eks"
>             groups = [
>               "system:masters"
>             ]
>         }]
> }
> ```
> 
> I think that should pretty much just work, but if somehow recerating the kubeconfig was triggered before adding that, they'll need to use a version of the kubeconfig without the role to edit the aws-auth configmap temporarily to add this so they can proceed.

Took care of this on this PR

PR #98, opened 2023-05-26
DOM-45710 • Enable access to dev-aws-eks-dataplane

adp312:
[DOM-45710](https://dominodatalab.atlassian.net/browse/DOM-45710)

A dataplane deployed with `dev-aws-eks-dataplane` deployer should be accessible via API, similar to `dev-cdk-eks-dataplane`.

Related PR: https://github.com/cerebrotech/platform-apps/pull/7348

### DIFF
--- a/submodules/eks/iam.tf
+++ b/submodules/eks/iam.tf
@@ -8,6 +8,15 @@ data "aws_iam_policy_document" "eks_cluster" {
       identifiers = ["eks.${local.dns_suffix}"]
     }
   }
+  statement {
+    sid     = "EKSClusterAssumeRole"
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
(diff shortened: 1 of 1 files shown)


submodules/eks/iam.tf
```diff
@@ -8,6 +8,15 @@ data "aws_iam_policy_document" "eks_cluster" {
       identifiers = ["eks.${local.dns_suffix}"]
     }
   }
+  statement {
+    sid     = "EKSClusterAssumeRole"
```
ddl-dclegg, line 12: We need a unique id here.

ddl-dclegg (changes requested):

submodules/eks/iam.tf
adp312: Good point. Agreed.

adp312:
Can someone review this please?

PR #103, opened 2023-06-07
DOM-43578: Add support for AWS IRSA.

ddl-czito:


submodules/eks/iam.tf
```diff
@@ -284,3 +284,22 @@ resource "aws_eks_identity_provider_config" "this" {
     username_prefix               = lookup(each.value, "username_prefix", null)
   }
 }
+
+data "aws_iam_policy_document" "kms" {
```
steved, line 288: Why does this need to be defined at the top level?

submodules/eks/variables.tf
```diff
@@ -195,6 +195,11 @@ variable "eks" {
       username_claim                = optional(string, null)
       username_prefix               = optional(string, null)
     })), [])
+    irsa = optional(object({
+      enabled                    = bool
+      namespace_service_accounts = list(string)
+      role_name                  = optional(string)
+    }))
```
steved, line 202: This structure assumes only a single role is necessary for Domino to function when we'll need to be more flexible over time, e.g.:
* backups will want a limited role that can only upload to S3
* EBS CSI driver will need a role that can manage volumes
* fluentd and nucleus actually have different requirements: fluentd only needs to read/write to S3 while nucleus needs S3 and EC2 API access.

submodules/irsa/main.tf
```diff
@@ -0,0 +1,61 @@
...
+      effect  = "Allow"
+      actions = ["sts:AssumeRoleWithWebIdentity"]
+
+      principals {
+        type        = "Federated"
+        identifiers = [each.value.provider_arn]
+      }
+
+      condition {
+        test     = endswith(statement.value, "*") ? "StringLike" : "StringEquals"
```
steved, line 36: If there aren't wildcards in the service account name are there any downsides to still using StringLike here?

steved (changes requested):
I think this is a good start and it's great you got it working, but I don't think it's quite ready to merge without us being able to answer how this extends to multiple different pods needing different permissions.

cc @miguelhar

PR #104, opened 2023-06-08
PLAT-6776 Adds RP delete perms

miguelhar:
[PLAT-6776](https://dominodatalab.atlassian.net/browse/PLAT-6776)

- Adds `backup:DeleteRecoveryPoint` perms
- Increment delay for `create_eks_role` assume(short term fix). Right fix is to have an intermediate step to ensure the role can be assumed. Will create follow up story


fraenkel (approved):
relying on time is always a slippery slope.

PR #106, opened 2023-06-22
DOM-47755 • Dataplane deployment failure

adp312:
https://dominodatalab.atlassian.net/browse/DOM-47755

Allow the data plane deployer to create an ACM certificate


iam-bootstrap/bootstrap-1.json
```diff
@@ -85,7 +85,11 @@
                 "ec2:RevokeSecurityGroupEgress",
                 "ec2:RevokeSecurityGroupIngress",
                 "ec2:RunInstances",
-                "ssm:GetParameter"
+                "ssm:GetParameter",
```
miguelhar, line 88: These should go under a different sid.

iam-bootstrap/bootstrap-1.json
```diff
@@ -91,6 +91,19 @@
...
+            "Sid": "ACMUngated",
+            "Effect": "Allow",
+            "Action": [
+                "acm:RequestCertificate",
+                "acm:DescribeCertificate",
+                "acm:GetCertificate",
+                "acm:ListTagsForCertificate"
+            ],
+            "Resource": [
+                "*"
```
miguelhar, line 104: Do we need * or can we restrict to `"arn:${partition}:acm:*:${account_id}:certificate/*"`?

iam-bootstrap/bootstrap-1.json
```diff
@@ -91,6 +91,19 @@
...
             ]
         },
+        {
+            "Sid": "ACMUngated",
+            "Effect": "Allow",
+            "Action": [
+                "acm:RequestCertificate",
+                "acm:DescribeCertificate",
+                "acm:GetCertificate",
+                "acm:ListTagsForCertificate"
```
miguelhar, line 101: Dont we need [DeleteCertificate](https://docs.aws.amazon.com/acm/latest/APIReference/API_DeleteCertificate.html)

miguelhar (changes requested):

iam-bootstrap/bootstrap-1.json
adp312: It depends on how paranoid we are. (Also, I suspect that `RequestCertificate` and `ListCertificates` will need `*`).

If you think we need restrict it by resource, I can do this, but will need another round of testing.

miguelhar:
Gonna add a note based on the teams past experience with acm. Also taking into account my ignorance on what the purpose is for this use case.
```
Sometime in the past, maybe 2018 or 2019, we'd been using ACM certs created dynamically at install time with our test installs. We did this so we could have the wildcard cert be on the subdomain with workspaces (they did workspace-xyz.deploy.domino.tech at the time rather than workspace-xyz-deploy.domino.tech like we do now), and this old schema required a dynamically created cert.
One day, cert creation broke for our test installs on one of the sandboxes. I made an AWS support case, and they told me that we'd been making too many certs, that they have to pay for each one they issue despite it being free to us, and we needed to stop provisioning throwaway ACM SSL certs as part of our CI. This was why we changed the workspace schema to workspace-xyz-deploy.domno.tech, so we could use a single wildcard cert per-sandbox-az, and not require this dynamic cert anymore.
I should note that we also tried to use Let's Encrypt in a similar fashion, and they started ratelimiting us.
I think for anything where we're tempted to dynamically create certs, we should either find a way around doing so (similar to what we did with workspaces) or bite the bullet and make a Domino CA of some sort that we can issue against, as the services generating commercial certs for this seem to view this sort of throwaway usage as problematic.
```

adp312:
So, you suggest not to create ACM certs in any automated fashion?

miguelhar:
> So, you suggest not to create ACM certs in any automated fashion?
> 
> https://dominodatalab.atlassian.net/browse/DOM-48092

Based on our tribal experience, ACM provisions are discouraged.

iam-bootstrap/bootstrap-1.json
miguelhar: thats fine we can keep `*`

miguelhar (approved):

PR #110, opened 2023-07-12
[DOM-48096] Adding support for VPC endpoint services

ldebello-ddl:

Private Link Requirements:

* Use NLB
* Use VPC Endpoint Services

Notes:


submodules/network/variables.tf
```diff
@@ -18,6 +18,16 @@ variable "region" {
   }
 }
 
+variable "vpc_endpoint_services" {
```
miguelhar, line 21: this var should be part of the network dict https://github.com/dominodatalab/terraform-aws-eks/blob/main/variables.tf#L238

ldebello-ddl: thanks for taking a look, I will move it

ldebello-ddl: fixed, please let me know if this looks good, anyway I am still testing the e2e before converting the PR from draft to PR

submodules/eks/iam.tf
```diff
@@ -237,6 +262,310 @@ data "aws_iam_policy_document" "custom_eks_node_policy" {
   ]
 }
 
+# https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.5.4/docs/install/iam_policy.json
```
steved, line 265: Is it easier to just keep a copy of this file instead of translating it to HCL?

submodules/eks/iam.tf
```diff
@@ -19,6 +19,31 @@ data "aws_iam_policy_document" "eks_cluster" {
...
+data "aws_iam_policy_document" "load_balancer_controller" {
+  statement {
+    sid     = ""
+    effect  = "Allow"
+    actions = ["sts:AssumeRoleWithWebIdentity"]
+
+    condition {
+      test     = "StringEquals"
+      variable = "${replace(aws_iam_openid_connect_provider.oidc_provider.id, "/arn:.*:oidc-provider//", "")}:sub"
+      values   = ["system:serviceaccount:domino-platform:${var.deploy_id}-load-balancer-controller"]
```
steved, line 31: The hardcoded namespace is fine for us internally, but might pose a problem for anyone trying to use this off-the-shelf

steved (commented):
Any way we could make this its own self-contained module?

submodules/eks/iam.tf
ldebello-ddl: you are right, I checked and there is no namespace parameter being injected in the TF, should I add the parameter or it is not intended? because from infra perspective having the namespace it could be misleading because the people running the TF could be a different that the one installing Domino although in this case they should provide the same value.

to be honest not 100% sure if this should be living here but as far as I understand this is the only available place at least right now.

What should we do?

ldebello-ddl:
> Any way we could make this its own self-contained module?

Probably yes, I am adding resources in IAM and Networking modules, I could create a new module to contains all the resources. Although the bootstrap will continue adding the required permissions.

Do you want me to move this in a module in this repo or use a different one?

submodules/eks/iam.tf
ldebello-ddl: Having the json file will be easier I will check how to do it

submodules/eks/iam.tf
steved: @miguelhar @fraenkel any thoughts on how we might want to handle this in the future and how it could be handled here?

steved:
> > Any way we could make this its own self-contained module?
> 
> Probably yes, I am adding resources in IAM and Networking modules, I could create a new module to contains all the resources. Although the bootstrap will continue adding the required permissions.
> 
> Do you want me to move this in a module in this repo or use a different one?

Responded offline, but copying here. I think a module in this repo is fine, just wanted to see if we could make it more self-contained. I think it might have some use for all customers, but since we're not testing it holistically yet I'm a bit worried about the impact.

submodules/eks/iam.tf
fraenkel: Both AKS and GKE pass a map of namespace names which is required.
The more I think about this, the more I think the management of the workload identities would best be served by the agent. Pushing it to the TF requires one to always make sure you are using the TF of the domino version so you can pick up any changes that may occur for a given release. I hit this hard when we moved hephaestus to the compute namespace. I don't want to get into the complications of doing so since it pushes additional requirements when running the agent.

ldebello-ddl:
Submodule refactor: https://github.com/dominodatalab/terraform-aws-eks/pull/124

PR #111, opened 2023-07-14
PLAT-6823: Node pool,cluster, infra isolation

miguelhar:
[PLAT-6822](https://dominodatalab.atlassian.net/browse/PLAT-6822)

- Provisioning is now separated into `infra`,`cluster` and `nodes`
- CI implements state migration from legacy deploys
- Move CI execution to sh.


ddl-dclegg (approved):
I've been chewing on this review for a while now (the PR is rather large), but I think it looks good.

PR #112, opened 2023-07-17
PLAT-6938: Run scripts with bash

fraenkel:
Execute scripts using bash not as a command to bash. This avoids issues with file permissions getting lost along the way.

### DIFF
--- a/submodules/eks/k8s.tf
+++ b/submodules/eks/k8s.tf
@@ -22,7 +22,7 @@ resource "terraform_data" "run_k8s_pre_setup" {
   ]
 
   provisioner "local-exec" {
-    command     = "./${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
+    command     = "bash ./${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
     interpreter = ["bash", "-c"]
     working_dir = module.k8s_setup[0].resources_directory
   }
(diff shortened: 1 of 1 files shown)


ddl-dclegg (approved):

submodules/eks/k8s.tf
```diff
@@ -22,8 +22,8 @@ resource "terraform_data" "run_k8s_pre_setup" {
   ]
 
   provisioner "local-exec" {
-    command     = "./${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
-    interpreter = ["bash", "-c"]
+    command     = "${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
+    interpreter = ["bash"]
```
Secretions, line 26: The command is a single string no matter what, so you might need to do something like this:

```
    command     = "bash ${module.k8s_setup[0].filename} set_k8s_auth set_eniconfig"
    interpreter = ["bash", "-c"]
```

Or else this, even though it's weird:

```
    command     = "set_eniconfig"
    interpreter = ["bash", module.k8s_setup[0].filename, "set_k8s_auth"]
```

Interpreter is optional, so long as it's equivalent to `bash -c` you might be able to just omit it too.

fraenkel: yea, going through all the stupid options... garbage.... could just make it all interpreter too.

PR #116, opened 2023-08-01
PLAT-6790 default tags to asgs

puluanau:
The aws provider default tags do not propagate to the ASG's because they are not created by terraform. However we already tag those ASG's with various things to assist with other k8s aspects of the deployment.  

This just adds those default tags into the ASGs, as well as bringing any custom nodegroup tags down to them as well.  In most instances this will improve the expectation from external auditing/IT orgs that things are tagged consistently.


steved:
I don't think this is a problem here, but just a heads up we had some trouble previously where the tags would refresh and cause issues: https://github.com/dominodatalab/terraform-aws-eks/pull/41

puluanau:
Will have to handle this externally.

PR #119, opened 2023-08-18
Support UpdateNodegroupConfig in bootstrap IAM, ignore min_size for nodegroup scaling

steved:
* eks:UpdateNodegroupConfig is used to change nodegroup configuration in-place
* If nodegroup `min_size` is changed, we don't particularly care and can ignore it


ddl-dclegg (approved):

steved:
Upgrade failure is expected until resolved in #120.

PR #120, opened 2023-08-21
support providing additional, custom KMS key policies

steved:


examples/kms-additional-policy/README.md
```diff
@@ -0,0 +1,71 @@
+
+# Create additional EKS node_group.
```
miguelhar, line 2: Not sure this piece is intended .

kms.tf
```diff
@@ -4,6 +4,9 @@ locals {
 
 data "aws_iam_policy_document" "kms_key_global" {
   count = local.create_kms_key
+
+  override_policy_documents = compact([var.kms.additional_policies])
```
miguelhar, line 8: if we make `var.kms.additional_policies` a list
can we just pass it through like 
```hcl
data "aws_iam_policy_document" "combined" {
  override_policy_documents = [
    data.aws_iam_policy_document.policy_one.json,
    data.aws_iam_policy_document.policy_two.json,
    data.aws_iam_policy_document.policy_three.json
  ]
```

examples/kms-additional-policy/kms-policy.json
```diff
@@ -0,0 +1,12 @@
+{
+ "Version": "2012-10-17",
+ "Statement": [
+   {
+      "Sid": "AllowAll",
+      "Effect": "Allow",
+      "Principal": "*",
```
ddl-dclegg, line 7: Nothing functionally wrong, but I wonder if our example should be a wildcard allow.

ddl-dclegg (approved):

PR #124, opened 2023-08-29
Initial commit submodule

ldebello-ddl:

Private Link Requirements:

* Use NLB
* Use VPC Endpoint Services

Notes:

PR Feedback from: https://github.com/dominodatalab/terraform-aws-eks/pull/110


submodules/privatelink/iam.tf
```diff
@@ -0,0 +1,50 @@
+data "aws_iam_policy_document" "load_balancer_controller" {
+  statement {
+    sid     = ""
+    effect  = "Allow"
+    actions = ["sts:AssumeRoleWithWebIdentity"]
+
+    condition {
+      test     = "StringEquals"
+      variable = "${replace(var.oidc_provider_id, "/arn:.*:oidc-provider//", "")}:sub"
```
steved, line 9: could store `replace(var.oidc_provider_id, "/arn:.*:oidc-provider//", "")` as a local

submodules/privatelink/outputs.tf
```diff
@@ -0,0 +1,6 @@
+output "info" {
+  description = "Target groups..."
```
steved, line 2: if you can, please fill this in before merging

submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -0,0 +1,153 @@
...
+
+  tags = {
+    "Name" = "${var.deploy_id}-${each.key}"
+  }
+}
+
+resource "aws_route53_record" "service_endpoint_private_dns_verification" {
+  for_each = local.endpoint_services
+
+  zone_id = data.aws_route53_zone.hosted[0].zone_id
```
steved, line 146: I think this assumes `route53_hosted_zone_name` is non-null (which is fine, but maybe just make the variable required?)

submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -0,0 +1,153 @@
+locals {
+  endpoint_services = tomap({ for service in var.vpc_endpoint_services : service.name => service.private_dns })
```
steved, line 2: is this `tomap` needed?

submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    principals {
+      type        = "Service"
+      identifiers = ["delivery.logs.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_s3_bucket" "lb_logs" {
```
steved, line 77: Can we reuse our existing monitoring bucket instead? Just keeps things in one place for now and infosec can just pull from it

steved (commented):
couple small comments

submodules/privatelink/outputs.tf
ldebello-ddl: it only return target groups, I added "..." based on other outputs that use the same pattern, I can remove it if you refer to that

submodules/privatelink/iam.tf
ldebello-ddl: fixed

submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: fixed

submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: making it a map it is easier to work in the tf

ldebello-ddl: any suggestion¿

submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: yes probably I will replace it

submodules/privatelink/vpc-endpoint-services.tf
steved: Oh I just wasn't sure why the tomap was needed I thought the map was created with
```
endpoint_services = { for service in var.vpc_endpoint_services : service.name => service.private_dns }
```

ldebello-ddl: let me check

PR #125, opened 2023-08-31
CDK to TF migration supporting changes

miguelhar:
### JIRA: [PLAT-7142](https://dominodatalab.atlassian.net/browse/PLAT-7142)
Changes supporting: [cdk-cf-eks](https://github.com/dominodatalab/cdk-cf-eks/pull/146)

* Documentation updates
* Update `eks` and `nodes` (example mods):
  * Take variables in appropriate precendence instead of merge.
  * Expose entire vars(`eks`, `default_node_groups` and `additional_node_groups`).
  * Expose `kms_info` on `eks`.
* Enhanced `tf.sh`:
  * Takes arbitrary `component`.
  * Adds support for creating and applying terraform plans.


examples/deploy/terraform/infra/variables.tf
```diff
@@ -153,6 +247,7 @@ variable "kms" {
 
 variable "eks" {
   description = <<EOF
+    creation_role_name = Only meant to support an imported role.
```
ddl-dclegg, line 250: This isn't quite clear to me. Maybe we should be more descriptive, like `Name of the role to import` or `Role with adequate migration permissions`?

examples/deploy/terraform/cluster/variables.tf
```diff
@@ -1,7 +1,79 @@
...
+    k8s_version = EKS cluster k8s version.
+    kubeconfig = {
+      extra_args = Optional extra args when generating kubeconfig.
+      path       = Fully qualified path name to write the kubeconfig file.
+    }
+    public_access = {
+      enabled = Enable EKS API public endpoint.
+      cidrs   = List of CIDR ranges permitted for accessing the EKS public endpoint.
+    }
+    "Custom role maps for aws auth configmap
```
ddl-dclegg, line 15: I think we're either missing the end quote or needing to change the beginning quote to a `#` (or remove it)

examples/deploy/terraform/cluster/variables.tf
```diff
@@ -1,7 +1,79 @@
...
+      username_prefix               = optional(string)
+    })))
+  })
+
+  default = null
+}
+
+variable "kms_info" {
+  description = <<EOF
+    Overrides the KMS key information. Meant for migrated configurations.
```
ddl-dclegg, line 68: This is kind of like the earlier line, where it's not a functional part of the map, but the proceeding values are, so maybe we should put a `#` before this

ddl-dclegg (approved):
Looks pretty good to me, I just have a few non-blocking comments to consider.

PR #126, opened 2023-08-31
[DOM-49677] Initial commit submodule

ldebello-ddl:

Private Link Requirements:

* Use NLB
* Use VPC Endpoint Services

Notes:

PR Feedback I: https://github.com/dominodatalab/terraform-aws-eks/pull/110
PR Feedback II: https://github.com/dominodatalab/terraform-aws-eks/pull/124


modules/eks/privatelink.tf
```diff
@@ -0,0 +1,10 @@
+module "privatelink" {
+  source                   = "./submodules/privatelink"
```
miguelhar, line 2: We need a toggle. this should be off by default. most consumers of the module will not use this. should be driven by a property in the `eks` var

tests/plan/terraform/main.tf
```diff
@@ -30,6 +30,7 @@ module "eks" {
   bastion_info        = module.infra.bastion
   create_eks_role_arn = module.infra.create_eks_role_arn
   tags                = module.infra.tags
+  monitoring_bucket   = module.infra.monitoring_bucket
```
miguelhar, line 33: We should create a plan file in `examples/tfvars` to test `private-link=ON`

miguelhar (changes requested):

modules/eks/privatelink.tf
ldebello-ddl: added

tests/plan/terraform/main.tf
ldebello-ddl: added

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -0,0 +1,85 @@
...
+
+  tags = {
+    "Name" = "${var.deploy_id}-${each.key}"
+  }
+}
+
+resource "aws_route53_record" "service_endpoint_private_dns_verification" {
+  for_each = local.endpoint_services
+
+  zone_id = data.aws_route53_zone.hosted[0].zone_id
```
steved, line 78: I still think this will fail without a route53_hosted_zone_name?

modules/eks/variables.tf
```diff
@@ -199,3 +199,41 @@ variable "tags" {
...
+  type        = string
+  description = "Monitoring bucket"
+  nullable    = false
+}
+
+variable "enabled_private_link" {
+  type        = bool
+  description = "Enable Private Link connections"
+  default     = false
+}
```
steved, line 239: Just a nitpick, but I think `enable_` instead of `enabled`

modules/eks/variables.tf
```diff
@@ -199,3 +199,41 @@ variable "tags" {
...
   default     = {}
 }
+
+variable "vpc_endpoint_services" {
+  description = <<EOF
+    [{
+      name      = Name of the VPC Enpoint Service.
+      ports     = List of ports exposing the VPC Enpoint Service. i.e [8080, 8081]
+      cert_arn  = Certificate ARN used by the NLB associated for the given VPC Endpoint Service.
+      private_dns = Private DNS for the VPC Enpoint Service.
```
steved, line 209:
```suggestion
      name      = Name of the VPC Endpoint Service.
      ports     = List of ports exposing the VPC Endpoint Service. i.e [8080, 8081]
      cert_arn  = Certificate ARN used by the NLB associated for the given VPC Endpoint Service.
      private_dns = Private DNS for the VPC Endpoint Service.
```

modules/eks/variables.tf
```diff
@@ -199,3 +199,41 @@ variable "tags" {
   description = "Deployment tags."
   default     = {}
 }
+
```
miguelhar, line 202: Dont we also need to pass through the `namespace` variable.

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -0,0 +1,85 @@
...
+  ]))
+}
+
+data "aws_route53_zone" "hosted" {
+  count        = var.route53_hosted_zone_name != null ? 1 : 0
+  name         = var.route53_hosted_zone_name
+  private_zone = false
+}
+
+resource "aws_lb" "nlbs" {
```
steved, line 21: I guess this is a broader question, but do we need a single NLB per vpce or can we have one shared NLB and vpce for all endpoints?

steved (approved):

modules/eks/submodules/privatelink/variables.tf
```diff
@@ -0,0 +1,98 @@
...
+variable "namespace" {
+  type        = string
+  description = "Namespace for IAM Policy conditions"
+  default     = "domino-platform"
+}
+
+variable "monitoring_bucket" {
+  type        = string
+  description = "Monitoring bucket"
+  nullable    = false
```
miguelhar, line 97: Is this a hard requirement even  when private_link = false?

miguelhar (changes requested):

modules/eks/submodules/privatelink/iam.tf
```diff
@@ -0,0 +1,54 @@
...
+      partition = data.aws_partition.current.partition
+    })
+  ]
+}
+
+resource "aws_iam_role" "load_balancer_controller" {
+  name               = "${var.deploy_id}-load-balancer-controller"
+  assume_role_policy = data.aws_iam_policy_document.load_balancer_controller.json
+  lifecycle {
+    ignore_changes = [name]
```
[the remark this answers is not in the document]
ldebello-ddl: probably a copy and past from another role, the name depends on the deploy_id that should not change otherwise we have other issues I will remove it

modules/eks/variables.tf
ldebello-ddl: fixed

modules/eks/variables.tf
ldebello-ddl: fixed

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: currently amazon does not allow to select the same NLB for multiples VPC endpoint services

modules/eks/variables.tf
ldebello-ddl: yes, adding it

modules/eks/submodules/privatelink/variables.tf
ldebello-ddl: removing the nullable and using default null

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: yes, not sure how to handle it, because I cannot use count with for_each, another options is make local.endpoint_services depend on var. route53_hosted_zone_name but sound hacky more than a solution, suggestions are welcome

modules/eks/submodules/privatelink/variables.tf
```diff
@@ -0,0 +1,97 @@
...
+
+  type = list(object({
+    name        = optional(string)
+    ports       = optional(list(number))
+    cert_arn    = optional(string)
+    private_dns = optional(string)
+  }))
+}
+
+variable "route53_hosted_zone_name" {
```
miguelhar, line 39: Given that this variable is `required` why are we toggling later on?
```
var.route53_hosted_zone_name != null ? 1 : 0
```

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
miguelhar: if you want to turn the resource off based on the presence of `var.route53_hosted_zone_name`, I think you can do :

```
for_each = var.route53_hosted_zone_name != null ? local.endpoint_services : {}
```

miguelhar (changes requested):

modules/eks/submodules/privatelink/variables.tf
ldebello-ddl: it is required for the private link module, but optional for the eks, it could receive null

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: moving to single variable and validation

modules/eks/submodules/privatelink/variables.tf
ldebello-ddl: fixing moving to object variable with validation

examples/deploy/terraform/cluster/main.tf
```diff
@@ -28,7 +28,10 @@ module "eks" {
   bastion_info        = local.infra.bastion
   create_eks_role_arn = local.infra.create_eks_role_arn
   tags                = local.infra.tags
-
+  privatelink = {
+    enabled           = false
```
miguelhar, line 32: should this be 
`enabled = var.enable_private_link`

miguelhar: shoudnt `route53_hosted_zone_name` be part of the `privatelink` so all `privatelink` vars be validated

modules/eks/privatelink.tf
```diff
@@ -0,0 +1,12 @@
+module "privatelink" {
+  count                    = var.privatelink.enabled ? 1 : 0
+  source                   = "./submodules/privatelink"
+  deploy_id                = var.deploy_id
+  region                   = var.region
+  vpc_endpoint_services    = var.privatelink.vpc_endpoint_services
```
miguelhar, line 6: could be consolidate and just pass `privatelink = var.privatelink`

modules/eks/variables.tf
```diff
@@ -199,3 +199,41 @@ variable "tags" {
...
+variable "privatelink" {
```
miguelhar, line 203: this var should be replicated in `modules/eks/submodules/privatelink/variables.tf` see how `eks` var is passed

miguelhar (changes requested):
`privatelink` var should be a dict from top level and just be passed down stream.

examples/deploy/terraform/cluster/main.tf
ldebello-ddl: I didnt get this, if I start adding the env vars I am replicating the same but in examples, is that the intention?

modules/eks/variables.tf
ldebello-ddl: done

modules/eks/privatelink.tf
ldebello-ddl: done

examples/deploy/terraform/cluster/main.tf
miguelhar: My mistake, what I meant here you covered elsewhere.  Given that this example is meant to be consume by external users and private link is off by default. I think we should omit this entry here.

ldebello-ddl: fixed

miguelhar (approved):

PR #140, opened 2023-09-21
PLAT-7275 Fallback fixes

miguelhar:
[PLAT-7275](https://dominodatalab.atlassian.net/browse/PLAT-7275)
* Sometimes the executor does not have perms to chmod the private key(`/root/pvt-key.pem`).
* Adds a retry on open ssh tunnel.

### DIFF
--- a/modules/eks/submodules/k8s/templates/k8s-pre-setup.sh.tftpl
+++ b/modules/eks/submodules/k8s/templates/k8s-pre-setup.sh.tftpl
@@ -1,12 +1,29 @@
-#! /usr/bin/env bash
-
-set -ex
+#!/usr/bin/env bash
+set -euo pipefail
 
 source ${k8s_functions_sh_filename}
 
+open_ssh_tunnel() {
+  local max_retries=5
+  local sleep_time=10
+  local success=0
+  for ((i = 1; i <= max_retries && success == 0; i++)); do
+    open_ssh_tunnel_to_k8s_api && success=1
+    if [[ $success -eq 0 && $i -ne $max_retries ]]; then
+      echo "Attempt ${i} failed. Retrying in ${sleep_time} seconds..."
+      sleep $sleep_time
+    fi
+  done
+
+  if [[ $success -eq 0 ]]; then
+    echo "Failed to open SSH tunnel after $max_retries attempts. Exiting..."
+    exit 1
+  fi
+}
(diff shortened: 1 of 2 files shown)


fraenkel (commented):
Even if it lives in root, you are saying we aren't executing as root and therefore the permissions may also be incorrect and fail on the next line.

miguelhar:
> Even if it lives in root, you are saying we aren't executing as root and therefore the permissions may also be incorrect and fail on the next line.

Next line is a read which can/should work, what am i missing?

fraenkel:
ssh will fail if the permissions are too wide.

miguelhar:
> ssh will fail if the permissions are too wide. otherwise why are we even `chmod`ing.

thats is true but given that the permissions fall in the middle it will work.(440)

```
#cant change perms
$ chmod 400 test.pem
chmod: Unable to change file mode on test.pem: Operation not permitted

# but can ssh
 ssh -oIdentitiesOnly=yes -i test.pem -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no ec2-user@34.211.197.67 -- hostname
Warning: Permanently added '34.211.197.67' (ED25519) to the list of known hosts.

ip-10-0-0-29.us-west-2.compute.internal
```

We chmod to be overly cautious when we are the ones that create the key

modules/eks/submodules/k8s/templates/k8s-pre-setup.sh.tftpl
```diff
@@ -1,12 +1,29 @@
...
+      echo "Attempt $i failed. Retrying in $sleep_time seconds..."
+      sleep $sleep_time
+    fi
+  done
+
+  if [[ $success -eq 0 ]]; then
+    echo "Failed to open SSH tunnel after $max_retries attempts. Exiting..."
+    exit 1
+  fi
+}
```
fraenkel, line 22: seems a bit overly complicated
```
for i in $(seq 1 $max_retries); do
    open_ssh_tunnel_to_k8s_api && return 0
   [ "$i" = "$max_retries" ] && echo "Failed to open ....." && return 1
  echo "Attempt $i failed. Retrying in $sleep_time seconds..."
  sleep "$sleep_time"
done
```

fraenkel (approved):

PR #141, opened 2023-09-22
Adds single-node module

miguelhar:
Adds single node module which can be used to provision an eks cluster with just one node instead of node_groups.


steved (approved):

modules/eks/submodules/k8s/templates/aws-auth.yaml.tftpl
```diff
@@ -11,6 +11,7 @@ data:
       groups:
         - system:bootstrappers
         - system:nodes
+        %{ if nodes_master }- system:masters%{ endif }
 %{ endfor ~}
 %{ for role, arn in eks_master_role_arns ~}
```
steved, line 16: Could we pass eks_custom_role_maps through instead?

steved (approved):

modules/eks/submodules/k8s/templates/aws-auth.yaml.tftpl
miguelhar: we need the nodes arn which gets created in the module.

PR #142, opened 2023-09-24
[DOM-46192] Adding required labels for cross account backups

ldebello-ddl:
Adding required labels to backup specific resources in a different account.

**Note:** I am not removing the existing aws_backup_vault because that will remove the recovery points, we should do that in a breaking change, also given the fact the new module is per account and this TF is per deployment that makes it tricky, at least we should wait until all Domino Cloud deployments are using the new module to avoid deleting possible backups

Backup Module: https://github.com/dominodatalab/terraform-aws-domino-backup


miguelhar:
Rather than harcoding this tag why dont we just pass it from [here](https://github.com/cerebrotech/platform-apps/blob/develop/src/deployer/terraform/generators/eks.py#L76)

ldebello-ddl:
> Rather than harcoding this tag why dont we just pass it from [here](https://github.com/cerebrotech/platform-apps/blob/develop/src/deployer/terraform/generators/eks.py#L76)

As far as I understand those tags are used as default tags for all the resources, for now we are just doing backup for EFS and two specific s3 buckets,.

miguelhar:
Im ok with the tag being in all components when it belongs to a deployment making use of it, but I dont think its suitable hardcoding the tag given that this module is consumed by several users and it would present on every deployment. Which is why i recommended conditionally setting it in the deployer.

ldebello-ddl:
> Im ok with the tag being in all components when it belongs to a deployment making use of it, but I dont think its suitable hardcoding the tag given that this module is consumed by several users and it would be present on every deployment. Which is why i recommended conditionally setting it in the deployer.

Yes, the issue of having the tag in all of components is the backup cross account is filtering by tag, similar discussion that I asked in https://dominodatalab.slack.com/archives/CBXBCC57B/p1694701908285579?thread_ts=1694701908.285579&cid=CBXBCC57B
 
I cannot configure stamp to backup specific resources because they do not exists when stamp is run (at least running stamp twice, which is not good for automating things), the other option is having half of the logic here and half in stamp.

Open to suggestion or other alternatives.

ldebello-ddl:
> Im ok with the tag being in all components when it belongs to a deployment making use of it, but I dont think its suitable hardcoding the tag given that this module is consumed by several users and it would be present on every deployment. Which is why i recommended conditionally setting it in the deployer.

Sharing backup module for reference https://github.com/dominodatalab/terraform-aws-domino-backup

ldebello-ddl:
> Im ok with the tag being in all components when it belongs to a deployment making use of it, but I dont think its suitable hardcoding the tag given that this module is consumed by several users and it would be present on every deployment. Which is why i recommended conditionally setting it in the deployer.

Other alternatives:

1. Add the tags for all the components and for domino cloud, disable backups for other resources than s3 and efs in thst case we will backup all s3 and efs having backup plan = remote, more buckets (we should validate how to disable backups resources per type I lknow there is a UI but not sure if that can be configure by terraform so we execute it from stamp.
2. Another option add something similar to the current change but add dynamic tags based ok resources type and name, we could inject from deployer something like {backup: {efs: [a], s3: [b, c, d]}} and then from tf merge tag for those componentes (hacky and tricky to implement, but could work)

miguelhar:
I think I understand the issue a bit better now, so the backup selection is filtering by tag and we need to sniper those we want included which are a subset of a type and more importantly specific deployments. I would be inclined to add a var in the `storage` submodule, `enable_remote_backup` which is then used in the module to conditionally set the tag in the module and it can be passed conditionally in the deployer.

steved (commented):
We probably also want the `logs` bucket and it might be a good idea to centralize the tag in a `local`, but LGTM

ldebello-ddl:
@miguelhar @steved I have added the required changes, please let me know if this looks good so I can start working on the deployer side

steved:
LGTM

ldebello-ddl:
@steved could you approve it?

README.md
```diff
@@ -281,3 +281,38 @@ Run the command below to generate a list of infrastructure values. These values
 ```
 
 This command will output a set of key-value pairs, extracted from the infrastructure setup, that can be used as inputs in the domino.yaml configuration file.
+
+
+## Domino Backups
+To make our backups safe we could use cross account backups (Accounts under same AWS Organization), in case you want to apply this you can use the following [module](https://github.com/dominodatalab/terraform-aws-domino-backup) to enabled this and the following steps:
```
steved, line 287: Are they required to be under the same org?
```suggestion
(Accounts under same AWS Organization)
```

README.md
```diff
@@ -281,3 +281,38 @@ Run the command below to generate a list of infrastructure values. These values
...
+│   ├── infra
+│   │   ├── main.tf
+```
+
+Content
+```
+provider "aws" {
+  alias   = "domino-backup"
+  region  = <<Backup Account Region>>
+  profile = <<Profile Account Here>>
```
steved, line 304: Maybe leave `profile` out since technically it could be any credential method?

README.md
```diff
@@ -281,3 +281,38 @@ Run the command below to generate a list of infrastructure values. These values
...
+}
+```
+
+2. Create a file called `backup.tf` in the infra module with the following content
+
+```
+module "backups" {
```
steved, line 311: Could we just have it all in a single file under `main.tf` just to simplify these instructions?

steved (approved):
Just some minor comments on wording

README.md
ldebello-ddl: yes, the accounts should be under the same aws organization (https://docs.aws.amazon.com/aws-backup/latest/devguide/create-cross-account-backup.html)

README.md
ldebello-ddl: fixed

README.md
ldebello-ddl: remove it

PR #143, opened 2023-09-25
fix typo

miguelhar:
[PLAT-7292](https://dominodatalab.atlassian.net/browse/PLAT-7292)


modules/infra/submodules/storage/s3.tf
```diff
@@ -455,7 +455,7 @@ resource "aws_s3_bucket_versioning" "buckets_versioning" {
   }
 }
 
-resource "aws_s3_bucket_public_access_block" "block_public_accss" {
+resource "aws_s3_bucket_public_access_block" "block_public_access" {
```
steved, line 458: maybe a `moved{}` block just to avoid churn?

steved (approved):

PR #150, opened 2023-09-29
DOM-50570 adds ecr:UntagResource

miguelhar:
[DOM-50570](https://dominodatalab.atlassian.net/browse/DOM-50570)


ddl-dclegg (approved):
Looks good

PR #153, opened 2023-10-04
add a note about deleting the VPC CNI addon for multiple jumps

steved:


fraenkel (approved):
just not sure anyone will see or read this

ddl-dclegg (approved):

PR #154, opened 2023-10-04
[DOM-47245] - domino-cost - kubecost long term storage

ddl-olsonJD:
### Link to JIRA
https://dominodatalab.atlassian.net/browse/DOM-47245

### What problem does this PR solve?
Kubecost currently requires a manual s3 setup for long-term storage, 

### What is the solution?
- automating s3 provisioning for kubecost long-term data storage.
- generate kubecost federated storage info need for kubernetes key creation.

reference
- https://github.com/cerebrotech/domino-config/pull/115


miguelhar (changes requested):

modules/infra/submodules/storage/main.tf
```diff
@@ -19,6 +19,13 @@ locals {
       policy_json = data.aws_iam_policy_document.blobs.json
       arn         = aws_s3_bucket.blobs.arn
     }
+    costs = {
+      count       = var.domino_cost.storage_enabled ? 1 : 0
```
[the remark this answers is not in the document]
ddl-olsonJD: thanks

modules/infra/submodules/storage/main.tf
```diff
@@ -19,6 +19,12 @@ locals {
       policy_json = data.aws_iam_policy_document.blobs.json
       arn         = aws_s3_bucket.blobs.arn
     }
+    costs = {
+      bucket_name = aws_s3_bucket.costs[0].bucket
```
miguelhar, line 23: This is not gonna work when the resource count is 0. this should be a conditional merge based on `var.domino_cost.storage_enabled`

modules/infra/submodules/storage/outputs.tf
```diff
@@ -31,5 +35,33 @@ output "info" {
...
     }
+    costs = {
+      storage_enabled = var.domino_cost.storage_enabled
+      cost_federated_store = yamlencode(
+        {
+          "type" : "S3",
+          "config" : {
+            "bucket" : aws_s3_bucket.costs[0].bucket,
+            "endpoint" : "s3.amazonaws.com",
+            "region" : "us-east-1",
```
miguelhar, line 46: hardcode?

modules/infra/submodules/storage/outputs.tf
```diff
@@ -31,5 +35,33 @@ output "info" {
...
+            },
+            "http_config" : {
+              "idle_conn_timeout" : "90s",
+              "response_header_timeout" : "2m",
+              "insecure_skip_verify" : false
+            },
+            "trace" : {
+              "enable" : true
+            },
+            "part_size" : 134217728
```
miguelhar, line 61: hardcode?

miguelhar (changes requested):

modules/infra/submodules/storage/outputs.tf
```diff
@@ -31,5 +35,9 @@ output "info" {
       container_registry = join("/", concat(slice(split("/", aws_ecr_repository.this["environment"].repository_url), 0, 1), [var.deploy_id]))
       iam_policy_arn     = aws_iam_policy.ecr.arn
     }
+    costs = {
+      storage_enabled = var.domino_cost.storage_enabled
+      costs_bucket    = aws_s3_bucket.costs[0].bucket
```
Secretions, line 40: I don't think this should need the separate output. There is an output that outputs all s3 buckets based on the local variable that's in `storage/main.tf`, so we can determine both the name of the costs bucket via that as well as whether it was provisioned by whether the value exists.

modules/infra/submodules/storage/dominocost.tf
```diff
@@ -0,0 +1,66 @@
+resource "aws_s3_bucket" "costs" {
```
Secretions, line 1: Since we have all the other s3 buckets centralized in s3.tf, I would think this should be there too. I don't think there's anything that makes it so this should be separated out vs the others.

modules/infra/submodules/storage/outputs.tf
ddl-olsonJD: thanks, will update.

ddl-olsonJD:
revalidated on this deployment jdbudget26409.amg-team-sandbox.domino.tech

.gitignore
```diff
@@ -50,3 +50,7 @@ eniconfig.yaml
 kubeconfig-proxy
 terraform.tfvars
 k8s-proxy-tunnel.sh
+
+# local files
+.DS_Store
+*/.idea/*
```
Secretions, line 56: Missing newline

Secretions (approved):

modules/infra/submodules/storage/variables.tf
```diff
@@ -120,3 +120,12 @@ variable "network_info" {
     })
   })
 }
+
+variable "domino_cost" {
+  description = "Determines whether to provision domino cost related infrastructures, ie, long term storage"
+  type = object({
+    storage_enabled = optional(bool, true)
```
miguelhar, line 127: Should the default be `false`?

miguelhar (changes requested):
Default value for domino_cost should be `false`.
We should add a plan test to `examples/tfvars` where we set it to `true`

miguelhar (approved):

PR #158, opened 2023-10-12
convert monitoring bucket away from ACLs

steved:
This is two changes in one:
1. We see (very) occasional timeouts when provisioning new infra: `(xyz-monitoring) ACL: operation error S3: PutBucketAcl, https response error StatusCode: 400`. My hunch is some sort of propagation delay for the ownership control setting.
2. We don't really need to use ACLs and the preference going forward is to just use policies. However, this is complicated because of our existing infrastructure. In order to set the policy to `BucketOwnerEnforced` the existing grants need to be removed. AFAICT for new installs, there's no way to automatically bypass the grant removal.

### DIFF
--- a/modules/infra/submodules/storage/s3.tf
+++ b/modules/infra/submodules/storage/s3.tf
@@ -335,14 +335,8 @@ resource "terraform_data" "set_monitoring_private_acl" {
       update_bucket() {
         ownership=$(aws s3api get-bucket-ownership-controls --bucket $bucket --output text --query "OwnershipControls.Rules[0]")
         if [[ "$ownership" == "BucketOwnerPreferred" ]] || [[ -z "$ownership" ]]; then
-          if ! aws s3api put-bucket-acl --bucket $bucket --acl private; then
-            return 1
-          fi
-
-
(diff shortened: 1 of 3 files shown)


modules/infra/submodules/storage/s3.tf
```diff
@@ -301,55 +324,51 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "monitoring" {
...
-resource "aws_s3_bucket_acl" "monitoring" {
-  depends_on = [aws_s3_bucket_ownership_controls.monitoring]
+      sleep_duration=10
+      bucket="${aws_s3_bucket.monitoring.bucket}"
 
-  bucket = aws_s3_bucket.monitoring.id
+      update_bucket() {
+        ownership=$(aws s3api get-bucket-ownership-controls --bucket $bucket --output text --query "OwnershipControls.Rules[0]")
+        if [[ "$ownership" == "BucketOwnerPreferred" ]] || [[ -z "$ownership" ]]; then
+          if ! aws s3api put-bucket-acl --bucket $bucket --acl private; then
```
miguelhar, line 338: maybe:
```
aws s3api put-bucket-acl --bucket $bucket --acl private; then || return 1
```

miguelhar (approved):

PR #174, opened 2023-11-14
* Bump EKS node group update timeout to 2h

JadeHayes:
We've encountered an issue during the creation of EKS clusters where the operation times out after 60 minutes, throwing an error related to an `ExpiredTokenException.` The root cause seems to be related to the update timeout during the cluster creation process.

This issue has been causing intermittent disruptions, and after investigation, we are proposing an update to the timeout settings related to EKS cluster node group creation


steved (approved):

fraenkel (changes requested):
this does not fix the issue.
see https://registry.terraform.io/providers/hashicorp/aws/latest/docs#duration

ddl-dclegg (approved):

PR #183, opened 2023-11-29
PLAT-7660 Adds default gpu labels/taints

miguelhar:
[PLAT-7660](https://dominodatalab.atlassian.net/browse/PLAT-7660)

If a `node_group` contains an instance with `itype` gpu or `node_group.gpu : true` add default labels and taints.


Secretions (approved):

ldebello-ddl (approved):
LGTM

PR #188, opened 2023-12-11
DOM-52659 Provision bucket for flyte

noahjax:
### Description
We want to provision an s3 bucket for flyte to use for pipeline results storage

### Jira
https://dominodatalab.atlassian.net/browse/DOM-52659


examples/tfvars/single-node.tfvars
```diff
@@ -38,4 +38,5 @@ storage = {
     force_destroy_on_deletion = true
   },
   costs_enabled = false
+  flyte_enabled = true
```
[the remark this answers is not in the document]
noahjax: Hmm yeah that seems like a good idea, let me give that a shot

ddl-risong-na (approved):

modules/infra/variables.tf
```diff
@@ -419,3 +419,10 @@ variable "domino_cur" {
 
   default = {}
 }
+
+variable "flyte_enabled" {
+  description = "Determines whether to provision flyte related infra, i.e. s3 buckets"
+  type        = bool
+  nullable    = true
+  default     = true
```
miguelhar, line 427: New features should always default to false in this module(for most cases), in order to avoid unintentional drift.  why do we need `nullable = true`

modules/infra/submodules/storage/variables.tf
```diff
@@ -122,3 +122,10 @@ variable "network_info" {
     })
   })
 }
+
+variable "flyte_enabled" {
+  description = "Determines whether to provision flyte related infra, i.e. s3 buckets"
+  type        = bool
+  nullable    = true
+  default     = true
```
miguelhar, line 130: should default to off.

modules/infra/main.tf
```diff
@@ -24,11 +24,12 @@ module "cost_usage_report" {
...
-  deploy_id    = var.deploy_id
-  network_info = module.network.info
-  kms_info     = local.kms_info
-  storage      = var.storage
+  source        = "./submodules/storage"
+  deploy_id     = var.deploy_id
+  network_info  = module.network.info
+  kms_info      = local.kms_info
+  storage       = var.storage
+  flyte_enabled = var.flyte_enabled
```
miguelhar, line 32: this variable should be a key on the storage variable

modules/infra/submodules/storage/variables.tf
```diff
@@ -122,3 +122,10 @@ variable "network_info" {
     })
   })
 }
+
+variable "flyte_enabled" {
```
miguelhar, line 126: should be moved into the `storage` dict

miguelhar (changes requested):

modules/infra/submodules/storage/variables.tf
noahjax: I put this outside of storage because eventually we will need to use `flyte_enabled` in other areas as well, but those other changes are still up in the air and out of scope for this initial PR. The intention was that making it a top level key would save us some refactoring effort down the line

modules/infra/submodules/storage/main.tf
```diff
@@ -24,6 +24,18 @@ locals {
       policy_json = data.aws_iam_policy_document.costs[0].json
       arn         = aws_s3_bucket.costs[0].arn
     } : {}
+    flyte_metadata = var.flyte_enabled ? {
```
noahjax, line 27: @miguelhar It's now unclear to me if references here should be `var.flyte_enabled` or `var.storage.flyte_enabled`. If it's the former then it doesn't seem like we need the `flyte_enabled` key in the storage module at all. If it's the latter then I'm unsure how we use the value of `var.flyte_enabled` to set the value of `var.storage.flyte_enabled`.

modules/infra/submodules/storage/s3.tf
```diff
@@ -603,3 +603,141 @@ data "aws_iam_policy_document" "costs" {
     }
   }
 }
+
+resource "aws_s3_bucket" "flyte_metadata" {
+  count               = var.flyte_enabled ? 1 : 0
```
noahjax, line 608: @miguelhar Similar point here about `var.flyte_enabled` vs `var.storage.flyte_enabled`.

noahjax:
Me and @miguelhar discussed and decided that the best course of action was to make flyte an object at the top level instead of a single variable so we can add more fields to it in the future. IMO also alleviates some of the confusion with passing flyte_enabled around, as we are now treating flyte as it's own entity.

modules/infra/submodules/storage/main.tf
```diff
@@ -24,6 +24,18 @@ locals {
       policy_json = data.aws_iam_policy_document.costs[0].json
       arn         = aws_s3_bucket.costs[0].arn
     } : {}
+    flyte_metadata = var.storage.flyte.enabled ? {
```
miguelhar, line 27: `var.flyte.enabled` ?

modules/infra/submodules/storage/main.tf
```diff
@@ -24,6 +24,18 @@ locals {
...
+      bucket_name = aws_s3_bucket.flyte_metadata[0].bucket
+      id          = aws_s3_bucket.flyte_metadata[0].id
+      policy_json = data.aws_iam_policy_document.flyte_metadata[0].json
+      arn         = aws_s3_bucket.flyte_metadata[0].arn
+    } : {}
+    flyte_data = var.storage.flyte.enabled ? {
```
miguelhar, line 33: `var.flyte.enabled` ?

miguelhar (changes requested):

modules/infra/submodules/storage/main.tf
noahjax: Yeah sorry was running into issues on the platform apps side so I tried adding `storage` here as a sanity check. It didn't resolve things, so undoing this commit

miguelhar (approved):

PR #194, opened 2024-01-04
[DOM-53463] Fixing target group name

ldebello-ddl:
Error

```
2024-01-04 14:57:19,585 - Plan: 12 to add, 0 to change, 2 to destroy.
2024-01-04 14:57:19,585 - 
2024-01-04 14:57:19,585 - Error: "name" cannot begin with a hyphen
2024-01-04 14:57:19,585 - 
2024-01-04 14:57:19,585 -   with module.eks.module.privatelink[0].aws_lb_target_group.target_groups["rabbitmq-external.5672"],
2024-01-04 14:57:19,585 -   on /resources/deployer/deploys/cloud-seer-bio/terraform/cluster/modules/eks/modules/eks/submodules/privatelink/vpc-endpoint-services.tf line 40, in resource "aws_lb_target_group" "target_groups":
2024-01-04 14:57:19,585 -   40:   name     = substr("${var.deploy_id}-${each.value.service}-${each.value.port}", -32, -1)
2024-01-04 14:57:19,585 - 
2024-01-04 14:57:19,585 - 
2024-01-04 14:57:19,585 - Error: "name" cannot begin with a hyphen
```

Fix

Use a random generator, max length of name is 32 and deploy id is 15 using random length will be 23

### DIFF
--- a/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
+++ b/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
@@ -37,7 +37,7 @@ resource "aws_lb" "nlbs" {
 resource "aws_lb_target_group" "target_groups" {
   for_each = { for entry in local.listeners : "${entry.service}.${entry.port}" => entry }
 
-  name     = "${var.deploy_id}-${substr(each.value.service,0,10)}-${each.value.port}"
+  name     = "${var.deploy_id}-${substr(each.value.service,0,9)}-${each.value.port}"
   port     = 80 # Not used but required
   protocol = "TCP"
   vpc_id   = var.network_info.vpc_id
--- a/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
+++ b/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
@@ -12,7 +12,7 @@ locals {
   ]))
 }
 
-resource "random_string" "target_group_random_gen" {
+resource "random_string" "target_group_random_suffix" {
   length  = 8
   special = false
 }
@@ -42,7 +42,7 @@ resource "aws_lb" "nlbs" {
 resource "aws_lb_target_group" "target_groups" {
   for_each = { for entry in local.listeners : "${entry.service}.${entry.port}" => entry }
 
-  name     = "${var.deploy_id}-${random_string.target_group_random_gen.result}"
+  name     = "${var.deploy_id}-${random_string.target_group_random_suffix.result}"
   port     = 80 # Not used but required
   protocol = "TCP"
   vpc_id   = var.network_info.vpc_id
--- a/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
+++ b/modules/eks/submodules/privatelink/vpc-endpoint-services.tf
@@ -12,6 +12,11 @@ locals {
   ]))
 }
 
+resource "random_string" "target_group_random_gen" {
+  length           = 8
+  special          = false
+}
+
 data "aws_route53_zone" "hosted" {
   name         = var.privatelink.route53_hosted_zone_name
   private_zone = false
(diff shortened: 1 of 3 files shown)


modules/eks/submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -37,13 +42,14 @@ resource "aws_lb" "nlbs" {
 resource "aws_lb_target_group" "target_groups" {
   for_each = { for entry in local.listeners : "${entry.service}.${entry.port}" => entry }
 
-  name     = substr("${var.deploy_id}-${each.value.service}-${each.value.port}", -32, -1)
+  name     = "${var.deploy_id}-${random_string.target_group_random_suffix.result}"
```
miguelhar, line 45: I believe the error is caused by the offset.

Could we just have:

```suggestion
  name     = substr("${var.deploy_id}-${each.value.service}-${each.value.port}", 0, 32)
```

miguelhar (changes requested):

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: I didnt want to do that because I can have repeat names, for example, for rabbitmq-external there two ports [ 5552,5672 ]

> substr("cloud-seer-bio-rabbitmq-external-5552", 0, 32)
"cloud-seer-bio-rabbitmq-external"
> substr("cloud-seer-bio-rabbitmq-external-5672", 0, 32)
"cloud-seer-bio-rabbitmq-external"

ldebello-ddl: that is why I choose to do it in reverse order but in that case during cloud-seer-bio that leads to a name starting with "-" so to avoid other issues changing implementation to <deploy-id> + random

and start with deploy-id is useful to search in aws ui

miguelhar: rather than a random string which offers no information, can we just do deploy_id+port, given that the ports are unique

miguelhar: Also can you expand on the duplicates, given that the deploy_id is 16 and this is 32, given that the string contains deploy_id + service +port, how can we have a duplicate with at least the deploy_id + 16 characters of the service?

ldebello-ddl: I want to avoid other to suffer the same issue in the future relying on the length of service name could be tricky it could work today but not tomorrow

ldebello-ddl: the existing service names are "vault-external" and "rabbitmq-external" but someone in the future could include a new service, so it could lead to the same issue

ldebello-ddl: regarding port today are different but it someone include another service with the same port that will fail

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -37,13 +37,14 @@ resource "aws_lb" "nlbs" {
 resource "aws_lb_target_group" "target_groups" {
   for_each = { for entry in local.listeners : "${entry.service}.${entry.port}" => entry }
 
-  name     = substr("${var.deploy_id}-${each.value.service}-${each.value.port}", -32, -1)
+  name     = "${var.deploy_id}-${substr(each.value.service,0,10)}-${each.value.port}"
```
miguelhar, line 40:
```suggestion
  name     = "${var.deploy_id}-${substr(each.value.service,0,9)}-${each.value.port}"
```
` (16 deploy_id + 5 port+ 2 - + 9 service)`

miguelhar (approved):
Appr

PR #197, opened 2024-01-09
DOM-52551 gov-cloud support, adds irsa module

miguelhar:
[DOM-52551](https://dominodatalab.atlassian.net/browse/DOM-52551)

* Adds IRSA module
The IRSA module contains a predefined configuration for `external-dns` as well as an entry for custom policies. Similar to `external-dns` can potentially add opinionated predefined polices like `cluster autoscaler` in the future.
In the case of `external-dns` we have the `use_cluster_odc_idp` toggle to control which IDP is used in the trust policy for the  IAM role. This is helpful in the case where we desired to have permissions for a particular hosted zone but it is not in the same account.For example a `us-gov` aws account(which have no public hosted zones) and a global(commercial) account that does, therefore allowing the configuration of a public hosted zone from the `us-gov` eks instance.


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ parameters:
     default: "v3.11.2"
   terraform_version:
     type: string
-    default: "1.4.4"
+    default: "1.6.6"
```
fraenkel, line 9: I am assuming we are updating TF in PA for 5.10?

examples/deploy/terraform/cluster/main.tf
```diff
@@ -25,13 +25,54 @@ module "eks" {
...
+  additional_irsa_configs = var.irsa_policies
 }
 
+# Provider configuration for the account where the hosted zone is defined.
+# Useful in configurations where accounts do not have a public hosted zone(i.e us-gov regions) and internet routing(public DNS)
+# is instead defined in a different account. Configure the `global` aws alias accordingly,
+# by specifying the profile belonging to the account pertaining to the hosted zone.
 provider "aws" {
-  region = var.infra.region
+  alias = "global"
```
fraenkel, line 69: this is a bit weird.
you are aliasing it as global but not setting it up based on the aws.global provider.

examples/deploy/terraform/cluster/variables.tf
```diff
@@ -82,3 +82,27 @@ variable "kms_info" {
...
+    policy              = string #json
+  }))
+
+  default = []
+}
+
+variable "irsa_external_dns" {
+  description = "Mappings for custom IRSA configurations."
+  type = object({
+    enabled             = optional(bool, false)
```
fraenkel, line 101: why have this if you could make irsa_external_dns nillable instead? it would also simplify all the checking logic.

modules/eks/submodules/irsa/variables.tf
```diff
@@ -0,0 +1,128 @@
+
+variable "eks_info" {
```
fraenkel, line 2: seems like a lot of detail.
one could pass the actual resources through or just query them as data, no?

modules/eks/submodules/irsa/variables.tf
```diff
@@ -0,0 +1,128 @@
...
+  EOF
+  type        = bool
+  default     = true
+}
+
+variable "external_dns" {
+  description = "Config to enable irsa for external-dns"
+
+  type = object({
+    enabled             = optional(bool, false)
```
fraenkel, line 104: nillable instead?

modules/infra/main.tf
```diff
@@ -100,7 +100,7 @@ locals {
 }
 
 provider "aws" {
-  region = "us-east-1"
+  region = strcontains(var.region, "us-gov") ? "us-gov-east-1" : "us-east-1"
```
fraenkel, line 103: this just seems strange and wrong given its for a specific purpose but the alias doesn't provide any inkling.
Does the service actually exist in gov cloud?

fraenkel (changes requested):
I realize some of this isn't your doing but we seem to prefer burying function/feature rather than composing at the top level. Its two ways of doing the same thing but it makes adding or tweaking harder since we cannot easily combine across modules unless the detail was pushed all the way down.
Just my $0.02.

examples/deploy/terraform/cluster/variables.tf
miguelhar: conditional on nillables cause issues, i.e bastion

modules/eks/submodules/irsa/variables.tf
miguelhar: we can not query the data if we want to provision in the global/commercial account. In addtion we treat the var as an object and just pass it around, rather than tailoring each variable.

modules/eks/submodules/irsa/variables.tf
miguelhar: same as before, I prefer to be explicit rather than condition based on null which causes issues like the bastion

modules/infra/main.tf
miguelhar: This service only exist in us-east-1 for commercial accounts, given the alias refers to that zone and we can not condition the provider.

examples/deploy/terraform/cluster/main.tf
miguelhar: we do not have the dual credentials set up, as a matter of fact we have not even discussed on how to do that yet. the idea is for the current session to be implicit(works for both gov/commercial) and the `global` to be explicit. If they are the same which is the case for commercial deployments then we toggle the `use_cluster_odc_idp` to `true` if they are not then `false` to create it using the `global` provider

miguelhar:
> I realize some of this isn't your doing but we seem to prefer burying function/feature rather than composing at the top level. Its two ways of doing the same thing but it makes adding or tweaking harder since we cannot easily combine across modules unless the detail was pushed all the way down. Just my $0.02.

Which is why `irsa` is called at a top level and its not a flag for the eks module, which will make the dual creds a bigger issue. this way a user can control it at the top level rather than making the eks module deal with the dual provider.

modules/eks/submodules/irsa/variables.tf
fraenkel: do we actually use all these values?
this should only require exactly what is needed, not what can easily be passed down.

modules/eks/submodules/irsa/variables.tf
fraenkel: I think its slightly different because of how the data is used and then later referenced. This detail is a leaf not something that is created and used elsewhere.

modules/infra/main.tf
fraenkel: us-east-1 and us-gov-east-1 are not the same.

fraenkel: what confuses me more is something line line 14. And that won't work for gov-cloud.

fraenkel:
It being top-level is good but the fact that its buried under eks makes it feel not so true as well as the variables require a lot of misc details that IRSA does not and should not need.

modules/eks/submodules/irsa/variables.tf
miguelhar: We are using the same approach everywhere, where the module outputs are treated like objects and passed around, I am not convinced there is an actual drawback to that, if it is then we will need to discuss it on its merits not as part of this.

modules/eks/submodules/irsa/variables.tf
miguelhar: I do not think nillables are preferred for this purpose in terraform, nor do I see what we gain.

modules/infra/main.tf
miguelhar: it wont work given it only exists in us-east-1, given that we need to give the provider something  I guess i could give it the `var.region` and short circuit the module so that its off in a gov region.
```hcl
module "cost_usage_report" {
  count        = var.domino_cur.provision_cost_usage_report && !strcontains(var.region, "us-gov") ? 1 : 0
```

fraenkel:
I believe our definitions of top-level differ.
The current IRSA requires you to use our EKS module rather than something that can be added regardless of how you deployed EKS.
This means anyone not using our module will have to translate the IRSA module to fit their use case.

fraenkel (approved):
I am not going to block this any further but I do think we will need to refine it as we better understand the requirements and use cases.

modules/infra/main.tf
miguelhar: I am not sure if its possible at least not straight forward, will bring it up with the CUR folks. for now I am just circuiting the module

modules/eks/submodules/irsa/variables.tf
miguelhar: After thinking about it more, what you said makes sense, given this could be used outside the module there is no reason to be buried, other than a lose relationship with eks.

PR #198, opened 2024-01-10
[DOM-53605] Adding security group to NLB to allow traffic to nodes

ldebello-ddl:
Nodes in target group were marked as Unhealthy, adding the security group to NLB and allowing traffic from NLB to eks-nodes fix it

also fixing an issue the NLB was public it should be internal because private link does not require that to be public


modules/eks/node-group.tf
```diff
@@ -51,6 +51,19 @@ resource "aws_security_group_rule" "node" {
   )
 }
 
+resource "aws_security_group_rule" "private_link" {
```
miguelhar, line 54: Instead of a new sgr can it be baked onto https://github.com/dominodatalab/terraform-aws-eks/blob/main/modules/eks/main.tf#L43

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
```diff
@@ -17,13 +17,31 @@ data "aws_route53_zone" "hosted" {
...
+  name        = "${var.deploy_id}-nlb-sg"
+  description = "NLB Security Group"
+  vpc_id      = var.network_info.vpc_id
+
+  egress {
+    description = "All traffic"
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
```
miguelhar, line 30: could we restrict this?

miguelhar (changes requested):

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
ldebello-ddl: I could use data resource to get CIDR from vpc_id and restrict it to VPC, do you think that makes sense?

modules/eks/node-group.tf
ldebello-ddl: I remember why I didnt use that, that rule is conditional for private link

miguelhar: I believe you can leverage the same conditional when adding it to `local.node_security_group_rules`,like this 
```hcl
node_security_group_rules = {
    private_link = var.privatelink.enabled ? {
        security_group_id        = aws_security_group.eks_nodes.id
        protocol                 = "tcp"
        from_port                = 30000
        to_port                  = 32767
        type                     = "ingress"
        description              = "NLB from private link to node ports"
        source_security_group_id = try(module.privatelink[0].info.nlb_sg, null)
    } : {}
  }
```

modules/eks/submodules/privatelink/vpc-endpoint-services.tf
miguelhar: the infra module should have this information, you can pass it through

modules/eks/node-group.tf
ldebello-ddl: I could use merge to add the condition object if you agree

miguelhar: yes, that should be fine. btw the idea is just to have these in one place

modules/eks/variables.tf
```diff
@@ -67,6 +67,7 @@ variable "network_info" {
         az_id     = string
       }))
     })
+    vpc_cidrs = string
```
miguelhar, line 70: are we missing `eips` here?

modules/eks/submodules/privatelink/variables.tf
```diff
@@ -42,6 +42,7 @@ variable "network_info" {
         az_id     = string
       }))
     })
+    vpc_cidrs = string
```
miguelhar, line 45: are we missing eips here?

miguelhar (approved):
I think we are missing `eips` in a couple of places.

modules/eks/variables.tf
ldebello-ddl: I didn't add it to reduce impact, but I can't include it if you want

PR #201, opened 2024-01-16
ignore duration tag

miguelhar:
[DOM-52985](https://dominodatalab.atlassian.net/browse/DOM-52985)

Ignore `duration` tag key


steved (approved):
Could be configurable, but I won't block it

fraenkel (approved):
I would have gone with `ignore_tags`. I get keys follows how its implemented.

PR #202, opened 2024-01-17
RE-1733 add SG rule for NewRelic metadata injector webhook

domino-cwellington:
Per RE-1649 we need an inbound SG rule. There needs to be a firewall rule open between the eks control plane and the deployment nodes for this to work.

This work is being tracked in RE-1733

### DIFF
--- a/modules/eks/main.tf
+++ b/modules/eks/main.tf
@@ -177,6 +177,14 @@ locals {
       type        = "egress"
       self        = true
     }
+    ingress_cluster_8443 = {
+      description                   = "Cluster API to NewRelic webhook"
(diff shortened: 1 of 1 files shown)


domino-cwellington:
I was going to try and have this rule added only when New Relic is enabled, however we're eventually going to switch this to the same port other webhooks use. Going to leave as is for now, since there's not much risk to this rule always being created.

domino-cwellington:
@dominodatalab/platform reviewer: We're primarily concerned that this is rolled out for Cloud deployments. Is this the flavor that gets used for those? Just want to confirm they will pick up this change.
Thanks.

PR #206, opened 2024-01-23
Update Flyte provisioning

noahjax:
### Description
This PR updates flyte provisioning to us to a state where we create a working Flyte deployment. This involves 2 key changes

1. Remove some restrictive s3 bucket policies that were blocking Flyte. We'll follow up later with more specific policies more tailored to Flyte.
2. Create irsa mappings for Flyte

### Jira
https://dominodatalab.atlassian.net/browse/DOM-52659
https://dominodatalab.atlassian.net/browse/DOM-53036


modules/irsa/outputs.tf
```diff
@@ -12,3 +12,14 @@ output "external_dns" {
     external_dns_use_eks_idp = var.use_cluster_odc_idp
   } : null
 }
+
+output "flyte" {
+  description = "Flyte info"
+  value = var.flyte.enabled ? {
+    eks = {
+      account_id            = regex("arn:.+:.+::(.*):.*", local.oidc_provider_arn)[0]
```
miguelhar, line 20: Nitpick
This seems a bit fragile, given that the account id can be the cluster or the one created here, we should pass in the the account id from the eks module, and use `data "aws_caller_identity" "current" {}` to the the id from the current session, use the similar logic to `local.oidc_provider_arn` to determine the value of a new variable `oidc_account`

miguelhar (changes requested):
Perhaps Im missing something, but where are the policies for `flyte_controlplane_role` and `flyte_dataplane_role`

PR #207, opened 2024-01-25
PLAT-7892 Adds a sh to handle multiple minors jump

miguelhar:
[PLAT-7892](https://dominodatalab.atlassian.net/browse/PLAT-7892)

Adds script to handle the eks `vpc-cni` addon upgrade in the case of more than 1 minor version jump.


miguelhar:
This is not needed anymore

PR #218, opened 2024-02-14
[DOM-54483] Link flyte roles to flyte buckets, add permissions

noahjax:
### Description
We need to add the permissions for flyte roles to allow them to access the relevant flyte s3 buckets

### Jira
https://dominodatalab.atlassian.net/browse/DOM-54483


modules/flyte/iam.tf
```diff
@@ -0,0 +1,102 @@
...
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Effect = "Allow"
+        Principal = {
+          Federated = local.oidc_provider_arn
+        }
+        Condition : {
+          StringEquals : {
+            "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
+            "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
+              "system:serviceaccount:domino-platform:flyteadmin",
```
miguelhar, line 16: I wouldnt hardcode the namespaces name nor the sa names. This should be defined in the deployer and passed through to both the terraform and later on to the installer.

modules/flyte/iam.tf
```diff
@@ -0,0 +1,102 @@
...
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Effect = "Allow"
+        Principal = {
+          Federated = local.oidc_provider_arn
+        }
+        Condition : {
+          StringEquals : {
+            "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
+            "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
+              "system:serviceaccount:domino-platform:datacatalog",
```
miguelhar, line 66: same as previous comment

modules/irsa/main.tf
```diff
@@ -1,10 +1,7 @@
-data "aws_caller_identity" "aws_account" {}
```
miguelhar, line 1: Why is this being removed?

modules/irsa/main.tf
```diff
@@ -1,10 +1,7 @@
...
-
 locals {
   oidc_provider_url = var.use_cluster_odc_idp ? var.eks_info.cluster.oidc.cert.url : aws_iam_openid_connect_provider.this[0].url
   oidc_provider_arn = var.use_cluster_odc_idp ? var.eks_info.cluster.oidc.arn : aws_iam_openid_connect_provider.this[0].arn
   name_prefix       = var.eks_info.cluster.specs.name
-  aws_account_id    = var.use_cluster_odc_idp ? var.eks_info.cluster.specs.account_id : data.aws_caller_identity.aws_account.account_id
```
miguelhar, line 7: Why is this being removed?

miguelhar (changes requested):

modules/flyte/iam.tf
noahjax: Yeah I was planning on addressing that in a separate PR since it is tracked by a different [ticket](https://dominodatalab.atlassian.net/browse/DOM-54326), but I can incorporate it here if you'd prefer

modules/irsa/main.tf
noahjax: I added [this](https://github.com/dominodatalab/terraform-aws-eks/pull/206/files#diff-82f0360a5fa9e26b1a605c21d229da63678b0727c5e6a13bde75b6ebdc60b7d2R1) for flyte when I initially put flyte stuff in irsa. Now that I am moving flyte stuff to it's own submodule it is no longer necessary.

modules/flyte/iam.tf
noahjax: [Here](https://github.com/dominodatalab/terraform-aws-eks/pull/221) is the other PR, I can merge it into this one if that makes sense

miguelhar: yes please.

modules/flyte/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "info" {
+  description = "Flyte info"
+  value = {
+    eks = {
+      metadata_bucket       = aws_s3_bucket.flyte_metadata.bucket
```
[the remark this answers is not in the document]
noahjax: There's a lot of value in keeping these buckets distinct so we can permission them more effectively

modules/flyte/iam.tf
```diff
@@ -0,0 +1,102 @@
...
+        Action = "sts:AssumeRoleWithWebIdentity"
+        Effect = "Allow"
+        Principal = {
+          Federated = local.oidc_provider_arn
+        }
+        Condition : {
+          StringEquals : {
+            "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
+            "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
+              "system:serviceaccount:${var.platform_namespace}:flyteadmin",
```
miguelhar, line 16: can we not hardcode the service accounts.

miguelhar: following the structure in https://github.com/dominodatalab/terraform-aws-eks/blob/main/modules/irsa/variables.tf#L70

modules/flyte/iam.tf
```diff
@@ -0,0 +1,102 @@
...
+        Effect = "Allow"
+        Principal = {
+          Federated = local.oidc_provider_arn
+        }
+        Condition : {
+          StringEquals : {
+            "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
+            "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
+              "system:serviceaccount:${var.platform_namespace}:datacatalog",
+              "system:serviceaccount:${var.compute_namespace}:*"
```
miguelhar, line 67: `*`? can we restrict to a service account?

miguelhar (changes requested):

modules/flyte/iam.tf
noahjax: We're still working on a better long term solution for this, but for now we need the wildcard. The issue is that we need runs to have permission to assume this role, and each run has it's own service account that is created dynamically.

modules/flyte/iam.tf
noahjax: Can you explain the value of making these configurable? These service account names are defined by Flyte and not likely to change between different installs

modules/flyte/iam.tf
noahjax: https://dominodatalab.atlassian.net/browse/DOM-54004

modules/flyte/iam.tf
miguelhar: yes, the value lies in the value originating outside the module in a place that it propagates to terraform and also to the installer-> charts. Failure to do so will cause a disconnect between the module and the installer given that they do not inherit the value.

miguelhar: Another approach is to actually use the variable :
https://github.com/dominodatalab/terraform-aws-eks/blob/63ada53bd6a6d380c45384ab86375b5fda4133b0/modules/flyte/variables.tf#L55
then set it as part of the outputs and have the installer pass it to the flyte chart

noahjax: I think I went with this "other approach" that you are suggesting...is there anything else you wanted me to do here?

modules/flyte/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "info" {
+  description = "Flyte info"
+  value = {
+    eks = {
```
miguelhar, line 4: Why does do we need this have a parent `eks`?

modules/flyte/iam.tf
miguelhar: I dont see the service accounts as part of the outputs, so I am assuming that the values will be passed from the deployer and populated to the installer.

modules/flyte/outputs.tf
noahjax: We wanted it to have a similar format to the domino config, which seems to have solidified around a similar level of nesting: https://github.com/cerebrotech/domino-config/blob/f3fc3f1af3520fee60f0d9aeb5f4c5df1b8237ae/domino_config/types1_7.py#L51-L52. Is this an issue that blocks the PR?

miguelhar (approved):

PR #225, opened 2024-03-05
[DOM-55148] Use KMS as default in Flyte s3 buckets

ddl-rliu:
Switch to KMS defaults from SSE-S3, for Flyte

See https://github.com/cerebrotech/platform-apps/pull/8821

https://dominodatalab.atlassian.net/browse/DOM-55148


ddl-rliu:
See https://github.com/dominodatalab/terraform-aws-eks/pull/226

PR #227, opened 2024-03-06
Minor flyte role updates

noahjax:
### Jira
https://dominodatalab.atlassian.net/browse/DOM-54483

### Description
Initial roles had a few minor issues that are cleaned up here


modules/flyte/iam.tf
```diff
@@ -13,7 +13,7 @@ resource "aws_iam_role" "flyte_controlplane" {
           StringEquals : {
             "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
             "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
-              "system:serviceaccount:${var.platform_namespace}:${var.serviceaccount_names.flyteadmin}",
+              "system:serviceaccount:${var.platform_namespace}:${var.serviceaccount_names.datacatalog}",
```
noahjax, line 16: Leaving datacatalog with controlplane role for now until we confirm that it needs data bucket access

modules/flyte/iam.tf
```diff
@@ -60,15 +60,29 @@ resource "aws_iam_role" "flyte_dataplane" {
           Federated = local.oidc_provider_arn
         }
         Condition : {
-          StringEquals : {
+          StringLike : {
```
noahjax, line 63: Oversight on my part in the initial PR, `StringEquals` doesn't work with the wildcard

modules/flyte/iam.tf
```diff
@@ -60,15 +60,29 @@ resource "aws_iam_role" "flyte_dataplane" {
...
             "${trimprefix(local.oidc_provider_url, "https://")}:aud" : "sts.amazonaws.com",
             "${trimprefix(local.oidc_provider_url, "https://")}:sub" : [
-              "system:serviceaccount:${var.platform_namespace}:${var.serviceaccount_names.datacatalog}",
               "system:serviceaccount:${var.compute_namespace}:*"
```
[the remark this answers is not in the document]
noahjax: Good point, no need to worry about supporting flyte native tasks for now

ddl-rliu (approved):

modules/flyte/iam.tf
noahjax: I think you're right here; input file uploads are just wonky ([issue](https://github.com/flyteorg/flyte/issues/5051))

Secretions (approved):

PR #233, opened 2024-03-15
PLAT-8122 Adds use_fips_endpoint var, default enabled for gov regions

miguelhar:
[PLAT-8122](https://dominodatalab.atlassian.net/browse/PLAT-8122)

Adds `use_fips_endpoints` option on modules.


modules/eks/versions.tf
```diff
@@ -39,4 +40,6 @@ provider "aws" {
   assume_role {
     role_arn = var.create_eks_role_arn
   }
+
+  use_fips_endpoint = strcontains(var.region, "us-gov") || var.use_fips_endpoints
```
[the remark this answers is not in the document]
miguelhar: Enforcing fips on us gov.

ddl-dclegg (approved):

examples/deploy/terraform/cluster/main.tf
```diff
@@ -44,7 +45,6 @@ locals {
 # the following annotation to the `external-dns` service account:
 # `eks.amazonaws.com/role-arn: <<module.irsa_external_dns.irsa_role>>`
 module "irsa_external_dns" {
-  count               = var.irsa_external_dns != null && var.irsa_external_dns.enabled ? 1 : 0
```
fraenkel, line 47: why did this change?

modules/eks/cluster.tf
```diff
@@ -104,6 +104,9 @@ resource "null_resource" "kubeconfig" {
   provisioner "local-exec" {
     when    = create
     command = "aws eks update-kubeconfig --kubeconfig ${self.triggers.kubeconfig_file} --region ${self.triggers.region} --name ${self.triggers.cluster_name} --alias ${self.triggers.cluster_name} ${local.kubeconfig.extra_args}"
+    environment = {
+      AWS_USE_FIPS_ENDPOINT = tostring(var.use_fips_endpoints)
+    }
   }
   provisioner "local-exec" {
```
fraenkel, line 111: shouldn't all the local-exec have the `AWS_USE_FIPS_ENDPOINT`?

modules/infra/versions.tf
```diff
@@ -28,4 +28,19 @@ provider "aws" {
...
+  region = strcontains(var.region, "us-gov") ? "us-gov-east-1" : "us-east-1"
+  alias  = "us-east-1"
+  default_tags {
+    tags = var.tags
+  }
+  ignore_tags {
+    keys = var.ignore_tags
+  }
+
+  use_fips_endpoint = var.use_fips_endpoints
```
fraenkel: why doesn't the above also have use_fips_endpoint?

examples/deploy/terraform/cluster/main.tf
miguelhar:
```
"The configuration of module.irsa_external_dns has its own local configuration for aws, and so it cannot accept an overridden configuration provided by the root module.",
```

Want to enforce the provider config `use_fips_endpoint` at the module level. Given the IRSA module does not/should not create anything without additional settings(i.e `var.irsa_external_dns.enabled` ). Not sure if it will be possible once integrated with how p-a consumes the module

modules/eks/cluster.tf
miguelhar: It doesnt hurt, but there are no aws calls on this one.

modules/infra/versions.tf
miguelhar: Not sure what you mean, both the default aws provider and the `"us-east-1"` alias have 
```
use_fips_endpoint = var.use_fips_endpoints
```

modules/irsa/variables.tf
```diff
@@ -66,7 +66,7 @@ variable "external_dns" {
 
   default = {}
   validation {
-    condition     = var.external_dns.enabled ? alltrue(var.external_dns.hosted_zone_name != null, var.external_dns.hosted_zone_name != "") : true
+    condition     = var.external_dns.enabled ? (var.external_dns.hosted_zone_name != null && var.external_dns.hosted_zone_name != "") : true
```
fraenkel, line 69:
```suggestion
    condition     = var.external_dns.enabled ? try(length(var.external_dns.hosted_zone_name), 0) > 0 : true
```

PR #237, opened 2024-04-04
update default instance types and volume throughput

steved:
* m7i-flex are slightly cheaper and offer better performance
* m6i are the same price as m5.2xlarge but don't have the downsides of requiring burst like m7i-flex (m7i are more expensive)
* 500MiB/s throughput was shown to have a significant impact on image pulls versus the additional cost


miguelhar (approved):

polovinko1980 (approved):
Good from scale CI/CD regression perspective

PR #243, opened 2024-04-11
[DOM-55975] Add CORS configuration for Flyte S3 buckets

ddl-rliu:
### Link to JIRA

[DOM-55975](https://dominodatalab.atlassian.net/browse/DOM-55975)

### What issue does this pull request solve?

Domino Flows UI cross-origin requests to Flyte S3 buckets fails.

### What is the solution?

Add CORS configuration for Flyte S3 buckets to allow these requests. The CORS configuration will allow origins that match the deployment url e.g. `https://flyte-testr42338.train-sandbox.domino.tech/`

https://github.com/cerebrotech/platform-apps/pull/8924 This PR adds the hostname to the terraform cluster config.


modules/flyte/s3.tf
```diff
@@ -106,3 +106,14 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "flyte_data_encryp
     ]
   }
 }
+
+resource "aws_s3_bucket_cors_configuration" "flyte_data" {
```
ddl-rliu, line 110: We've considered the alternative where instead of using CORS policy, instead use an iframe. However, the file renderer for Flows will make a fetch to the S3 bucket pre-signed URL, so the iframe does not seem suitable for the use case.

modules/flyte/s3.tf
```diff
@@ -106,3 +106,14 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "flyte_data_encryp
...
+  bucket = aws_s3_bucket.flyte_data.id
+
+  cors_rule {
+    allowed_headers = []
+    allowed_methods = ["GET"]
+    allowed_origins = [var.user_host]
```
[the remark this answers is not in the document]
ddl-rliu: I believe `domino.userHost` (that's the platform apps var that gets passed down into this config) does take into account vanity URL support.

Some relevant code: https://github.com/cerebrotech/platform-apps/blob/d840f75cc5e68764545b717abd9f357c267b3015/src/deployer/deployment_manifest.py#L242 `hostname = self.vanity_url if self.vanity_url else f"{self.deploy_id}.{root_domain}"`
```
"domino": {
    "userHost": f"https://{hostname}",
},
```

Still looking into the possibility of multiple names per domino deploy

modules/flyte/s3.tf
```diff
@@ -106,3 +106,14 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "flyte_data_encryp
...
   }
 }
+
+resource "aws_s3_bucket_cors_configuration" "flyte_data" {
+  bucket = aws_s3_bucket.flyte_data.id
+
+  cors_rule {
+    allowed_headers = []
+    allowed_methods = ["GET"]
+    allowed_origins = ["*"]
```
ddl-rliu, line 116: There is a potential issue that `"*"` is too lax. However, since the bucket is protected behind IAM roles, this concern is mitigated. For example, if an object is accessed in a non-domino domain (potentially bad), the associated pre-signed URL expires after 5 minutes anyways, mitigating this concern.

There is a potential alternative that uses nucleus-fronted nginx server to proxy these requests. [Joaquin's POC here](https://dominodatalab.slack.com/archives/CBXBCC57B/p1713218041510889?thread_ts=1712958787.781829&cid=CBXBCC57B). This would work around CORS for small (<5MB) files, and large files would use a download link instead (not affected by CORS). A concern with this approach is how it follows a "bad" pattern of proxying through Nucleus, since we may be moving to avoid this kind of pattern for artifact access, and instead preferring to have a more "direct" flow. 

---

tldr [summary of discussion that happened here](https://dominodatalab.slack.com/archives/CBXBCC57B/p1712958787781829).

Secretions (approved):

PR #245, opened 2024-04-12
pull through cache for calico on quay

steved:
In order to decrease the risks of a Quay.io outage, we can specify a pull-through cache as the upstream registry for Calico.


modules/eks/submodules/k8s/variables.tf
```diff
@@ -1,7 +1,7 @@
 variable "calico_version" {
   type        = string
   description = "Calico operator version."
-  default     = "v3.25.2"
```
steved, line 4: Wasn't totally sure if this was ok. It's not necessary for this change, but I figured while I'm updating things.

modules/eks/submodules/k8s/templates/k8s-functions.sh.tftpl
```diff
@@ -119,8 +119,9 @@ install_calico() {
       --namespace "tigera-operator" \
       --set installation.kubernetesProvider=EKS \
       --set installation.cni.type=AmazonVPC \
-      --set installation.registry="quay.io/" \
+      --set installation.registry="${calico_image_registry}/" \
       --set installation.fipsMode="${calico_fips_mode}" \
+      --wait \
```
steved, line 124: I also don't totally remember if not setting this was on purpose. I can imagine it is if say calico-node takes a while to rolling update

modules/eks/k8s.tf
```diff
@@ -12,6 +17,9 @@ module "k8s_setup" {
   eks_info          = local.eks_info
   use_fips_endpoint = var.use_fips_endpoint
 
+  # Note: must match modules/infra/submodules/storage/ecr.tf
```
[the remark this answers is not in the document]
miguelhar: Id prefer that, although recognize there is a bit of wiring involved.

modules/eks/submodules/k8s/variables.tf
miguelhar: only bumped the patch for the same reason, but I am not aware of any issues.

modules/infra/submodules/storage/iam.tf
```diff
@@ -56,6 +56,27 @@ data "aws_iam_policy_document" "ecr" {
...
+}
+
+data "aws_iam_policy_document" "ecr_pull_through_cache" {
+  count = local.supports_pull_through_cache ? 1 : 0
+
+  statement {
+    effect = "Allow"
+
+    resources = [
+      "arn:${data.aws_partition.current.partition}:ecr:${var.region}:${data.aws_caller_identity.this.account_id}:repository/${aws_ecr_pull_through_cache_rule.quay[0].ecr_repository_prefix}/*"
```
miguelhar, line 70: we have this starting in TF 1.8 
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/functions/arn_build

modules/nodes/variables.tf
```diff
@@ -101,7 +101,8 @@ variable "eks_info" {
     }
   EOF
   type = object({
-    k8s_pre_setup_sh_file = string
+    k8s_pre_setup_change_hash = string
```
miguelhar, line 104: These need to be optional, otherwise the plans will fail, due to the outputs not being updated for infra|eks at plan time.

modules/single-node/variables.tf
```diff
@@ -101,7 +101,8 @@ variable "eks_info" {
     }
   EOF
   type = object({
-    k8s_pre_setup_sh_file = string
+    k8s_pre_setup_change_hash = string
```
miguelhar, line 104: same

miguelhar (changes requested):

modules/infra/submodules/storage/iam.tf
steved: I'd prefer to wait a bit to not bundle this with the TF upgrade since there's a handful of other places we're doing this

modules/nodes/main.tf
```diff
@@ -66,7 +66,7 @@ resource "terraform_data" "calico_setup" {
   count = try(fileexists(var.eks_info.k8s_pre_setup_sh_file), false) ? 1 : 0
 
   triggers_replace = [
-    filemd5(var.eks_info.k8s_pre_setup_sh_file)
+    var.eks_info.k8s_pre_setup_change_hash
```
miguelhar, line 69: I would add an entry in `modules/eks/submodules/k8s/templates/k8s-functions.sh.tftpl`
`K8S_function_hash=md5(k8s-functions.sh`

then keep the trigger as `filemd5(var.eks_info.k8s_pre_setup_sh_file)`
I believe it would avoid a new var being passed around

miguelhar (approved):

modules/nodes/main.tf
steved: ah yeah, nice!

PR #251, opened 2024-06-17
[DOM-58654] - updated tagging permissions

ddl-olsonJD:
### problem
terraform upgrades is failing during upgrades when cloud_billing resources are avaiable due to missing permissions.

### solution
update deployment role to include tagging permissions.

### DIFF
--- a/modules/iam-bootstrap/bootstrap-0.json
+++ b/modules/iam-bootstrap/bootstrap-0.json
@@ -126,7 +126,9 @@
         "logs:ListTagsForResource",
         "logs:PutRetentionPolicy",
         "logs:TagLogGroup",
-        "logs:TagResource"
+        "logs:UntagLogGroup",
+        "logs:TagResource",
+        "logs:UntagResource"
       ],
       "Resource": "*"
     },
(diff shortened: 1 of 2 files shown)


modules/iam-bootstrap/bootstrap-0.json
```diff
@@ -125,7 +125,8 @@
         "logs:ListTagsLogGroup",
         "logs:ListTagsForResource",
         "logs:PutRetentionPolicy",
-        "logs:TagLogGroup"
+        "logs:TagLogGroup",
+        "logs:TagResource"
```
ddl-dclegg, line 129: This is fine, but I've been finding that I should've added Untag with a lot of these too, only after seeing an error down the line. Can we add `UntagLogGroup` and `UntagResource`? https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncloudwatchlogs.html

ddl-dclegg (approved):

modules/iam-bootstrap/bootstrap-0.json
ddl-olsonJD: of course.

PR #257, opened 2024-08-15
External Deployments Module

dmcwhorter-ddl:
Module to provide an IAM role and permissions for the model deployment operator service account to assume the role.


dmcwhorter-ddl:
will move toward a simpler approach that adds resources to irsa module (with a flag to turn on and off)

PR #258, opened 2024-08-19
External Deployments Operator IAM Role

dmcwhorter-ddl:
create external deployments iam role using new resource under existing irsa module


miguelhar:
@dmcwhorter-ddl can you add a plan test [here](https://github.com/dominodatalab/terraform-aws-eks/tree/main/examples/tfvars) with `external_deployments_operator` enabled

miguelhar (approved):

PR #267, opened 2024-09-10
External Deployments Role Policies

dmcwhorter-ddl:
Adds policies to allow deploying to sagemaker within the same AWS account


dmcwhorter-ddl:
@miguelhar This is working in my testing, could you give me some initial feedback on the changes?
Of note I have 3 outstanding issues I need to address:
1. Add a flag/option to enable/disable granting the policies to enable deployments in the AWS account
2. Add another flag/option to enable/disable granting the blanket assume role permission for other AWS accounts/role
3. Fix whatever is broken with the upgrade test

modules/irsa/external-deployments-operator.tf
```diff
@@ -18,6 +18,192 @@ resource "aws_iam_role" "external_deployments_operator" {
...
+      "ecr:GetAuthorizationToken",
+      "ecr:DescribeRegistry"
+    ]
+    resources = ["*"]
+  }
+  statement {
+    sid       = "KmsDecryptDominoBlobs"
+    effect    = "Allow"
+    actions   = ["kms:Decrypt"]
+    resources = [local.kms_key_arn]
```
miguelhar, line 85: does null work here ?
```
kms_key_arn = var.kms_info.enabled ? var.kms_info.key_arn : null
```

miguelhar (changes requested):

modules/irsa/external-deployments-operator.tf
dmcwhorter-ddl: will add a check, good call

examples/deploy/terraform/cluster/main.tf
```diff
@@ -7,14 +7,15 @@ data "terraform_remote_state" "infra" {
 }
 
 locals {
-  infra = data.terraform_remote_state.infra.outputs.infra
-  kms   = var.kms_info != null ? var.kms_info : local.infra.kms
+  infra  = data.terraform_remote_state.infra.outputs.infra
+  kms    = var.kms_info != null ? var.kms_info : local.infra.kms
+  region = var.region != null ? var.region : local.infra.region
```
miguelhar, line 12: I am not sure we want this here, given that we need to deploy in the region where infra was created

examples/deploy/terraform/cluster/main.tf
```diff
@@ -55,6 +56,9 @@ module "irsa_external_dns" {
   use_cluster_odc_idp = local.is_eks_account_same
   eks_info            = module.eks.info
   external_dns        = var.irsa_external_dns
+  region              = local.region
```
miguelhar, line 59: Why does this `irsa` module instance which is used to enable irsa_external_dns need these now.

examples/deploy/terraform/cluster/main.tf
```diff
@@ -71,6 +75,9 @@ module "irsa_policies" {
   use_cluster_odc_idp     = true
   eks_info                = module.eks.info
   additional_irsa_configs = var.irsa_policies
+  region                  = local.region
```
miguelhar, line 78: similar as previous comment

examples/deploy/terraform/cluster/main.tf
```diff
@@ -79,6 +86,9 @@ module "irsa_external_deployments_operator" {
   use_cluster_odc_idp           = local.is_eks_account_same
   eks_info                      = module.eks.info
   external_deployments_operator = var.irsa_external_deployments_operator
+  region                        = local.region
```
miguelhar, line 89: If this fields are needed by `external_deployments_operator` they should be contained in its definition and not be a requirement by the whole module.
see [this](https://github.com/dominodatalab/terraform-aws-eks/blob/main/modules/irsa/variables.tf#L105) example of the variable being passed

examples/deploy/terraform/cluster/main.tf
```diff
@@ -92,19 +102,26 @@ module "irsa_external_deployments_operator" {
...
   ignore_tags {
     keys = local.infra.ignore_tags
   }
   use_fips_endpoint = var.use_fips_endpoint
 }
 
 provider "aws" {
-  region = local.infra.region
+  region = local.region
+  default_tags {
```
miguelhar, line 116: Why do we need this now?

modules/irsa/variables.tf
```diff
@@ -88,13 +88,38 @@ variable "additional_irsa_configs" {
   }
 }
```
miguelhar, line 90: if `kms_info` `region` are needed by `external_deployments_operator` they should be fields and not required variables by the module

miguelhar (dismissed):

modules/irsa/external-deployments-operator.tf
```diff
@@ -1,23 +1,280 @@
...
+  assume_role_policy = data.aws_iam_policy_document.external_deployments_operator_assume_role_policy[0].json
+}
+
+data "aws_iam_policy_document" "external_deployments_decrypt_blobs_kms" {
+  count = var.external_deployments_operator.enabled ? 1 : 0
+  statement {
+    sid       = "KmsDecryptDominoBlobs"
+    effect    = "Allow"
+    actions   = ["kms:Decrypt"]
+    resources = [coalesce(var.external_deployments_operator.kms_key_arn, "ignored")]
```
steved, line 86: can you gate the `count` on kms_key_arn too?

tests/plan/terraform/versions.tf
```diff
@@ -12,7 +12,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5.0"
+      version = "<5.67.0"
```
steved, line 15: why is this pinned?

modules/irsa/external-deployments-operator.tf
```diff
@@ -1,23 +1,280 @@
...
+    condition {
+      test     = "StringEquals"
+      variable = "${trimprefix(local.oidc_provider_url, "https://")}:sub"
+      values = [
+        "system:serviceaccount:${var.external_deployments_operator.namespace}:${var.external_deployments_operator.service_account_name}"
+      ]
+    }
+  }
+}
+data "aws_iam_policy_document" "external_deployments_self_sagemaker_assume_role" {
```
steved, line 38: I don't totally follow, why doesn't the service account just have these permissions granted locally if `grant_in_account_policies`? Why the double hop?

modules/irsa/external-deployments-operator.tf
```diff
@@ -1,23 +1,280 @@
...
+      "ecr:GetDownloadUrlForLayer",
+    ]
+    resources = [
+      "arn:${data.aws_partition.current.partition}:ecr:${local.region}:${local.account_id}:repository/${local.environments_repository}",
+      "arn:${data.aws_partition.current.partition}:ecr:${local.region}:${local.account_id}:repository/${local.environments_repository}*"
+    ]
+  }
+}
+
+data "aws_iam_policy_document" "external_deployments_assume_any_role" {
```
steved, line 254: Why would this be used? Couldn't it just assume something with super broad access and do whatever?

steved (approved):
Mechanically, seems fine and if you've tested it feel free to merge.

In a vacuum, the external-deployments IAM code could probably be in its own module to clean up all the repeated `count = var.external_deployments_operator.enabled ...` statements.

I'm also not up to speed on the contents of the policies so I just left questions, but they're non-blocking.

tests/plan/terraform/versions.tf
dmcwhorter-ddl: There was a bug with the latest release that was breaking the build. Will find it and link

modules/irsa/external-deployments-operator.tf
dmcwhorter-ddl: This is used in two situations 
1. The operator is built to start with its own role and then assume another role possibly in a different AWS account 
2. When creating limited access short term credentials with a specific policy to hand back to users to allow them to invoke the endpoint or download logs

modules/irsa/external-deployments-operator.tf
dmcwhorter-ddl: This is here so that the operator role can assume a different role to deploy endpoints in a different AWS account.  The other accounts role also has to grant access to this role to assume it.  This is just giving the domino was account permission to assume other roles if the corresponding policy is in place on the other side.
I figure there could be situations when this is too broad so that was why there's a flag to allow it to be disabled.
Lmk if you have any ideas about a better way to do this.  The main goal is that you can setup to deploy endpoints in a different AWS account without having to make changes to the domino aws account

dmcwhorter-ddl:
Reworked in favor of the approach in https://github.com/dominodatalab/terraform-aws-eks/pull/272
closing this without merging

PR #272, opened 2024-09-16
External Deployments Module + Policy Options 

dmcwhorter-ddl:
1. Move External Deployments to a separate module
2. Add option to grant policies for 
    a. in-account deployment
    b. assuming external account roles


examples/deploy/terraform/cluster/variables.tf
```diff
@@ -118,13 +118,18 @@ variable "use_fips_endpoint" {
   default     = false
 }
 
-variable "irsa_external_deployments_operator" {
+variable "external_deployments_operator" {
```
steved, line 121: fwiw, keeping this variable structure as input to `modules/external-deployments` might also be preferable to flattening it as you pass it through

modules/external-deployments/operator_role_policies.tf
```diff
@@ -0,0 +1,205 @@
...
+  }
+}
+
+data "aws_iam_policy_document" "assume_any_role" {
+  statement {
+    sid       = "StsAllowOtherAccountsAssumeRole"
+    effect    = "Allow"
+    actions   = ["sts:AssumeRole"]
+    resources = ["*"]
+    condition {
```
dmcwhorter-ddl, line 180: @steved I did some testing and this restriction *does* keep the role from assuming another role in the same account.
I tested with the `smtest61590-eks` role.  When I removed this condition the `smtest61590-external-deployments-operator` role was able to assume `smtest61590`.  When I have this condition in here the assume role fails with
```
An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts::399181051250:assumed-role/smtest61590-external-deployments-operator/botocore-session-1726579049 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::399181051250:role/smtest61590-eks
```

dmcwhorter-ddl: So I think this should satisfy the concern about not being able to assume any other role in the same AWS account while still enabling this role to assume a role in another account without changes in this AWS account?

miguelhar (approved):

steved (approved):

PR #278, opened 2024-10-04
fix invalid type for identity providers config

steved:
`required_claims` is an invalid type and so identity providers cannot be configured.


miguelhar (approved):

miguelhar:
will need to re-run as aws fixed this.
```
InvalidParameterException: ami type AL2023_x86_64_NVIDIA is not available
```

PR #286, opened 2024-10-31
replace - with _ on volume names

miguelhar:
Avoids failure incase the deployid contains `-`

```
Error: creating FSx for NetApp ONTAP Volume (netapp-poc_domino_shared_storage): operation error FSx: CreateVolume, https response error StatusCode: 400, RequestID: 54d9520f-2f83-4cfc-b9b9-9e43b7874cdb, BadRequest: netapp-poc_domino_shared_storage is not a valid volume name. A volume name must begin with a letter or underscore, can only contain alphanumeric characters or "_", and cannot exceed 203 characters in length.
```

### DIFF
--- a/modules/infra/submodules/storage/netapp.tf
+++ b/modules/infra/submodules/storage/netapp.tf
@@ -154,7 +154,7 @@ resource "aws_fsx_ontap_storage_virtual_machine" "eks" {
 resource "aws_fsx_ontap_volume" "eks" {
   count                      = local.deploy_netapp && var.storage.netapp.volume.create ? 1 : 0
   storage_virtual_machine_id = aws_fsx_ontap_storage_virtual_machine.eks[0].id
-  name                       = replace("${var.deploy_id}_${var.storage.netapp.volume.name_suffix}", "/[^a-z_]/", "_")
+  name                       = replace("${var.deploy_id}_${var.storage.netapp.volume.name_suffix}", "/[^a-zA-z0-9_]/", "_")
   junction_path              = var.storage.netapp.volume.junction_path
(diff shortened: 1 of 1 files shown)


modules/infra/submodules/storage/netapp.tf
```diff
@@ -154,7 +154,7 @@ resource "aws_fsx_ontap_storage_virtual_machine" "eks" {
 resource "aws_fsx_ontap_volume" "eks" {
   count                      = local.deploy_netapp && var.storage.netapp.volume.create ? 1 : 0
   storage_virtual_machine_id = aws_fsx_ontap_storage_virtual_machine.eks[0].id
-  name                       = "${var.deploy_id}_${var.storage.netapp.volume.name_suffix}"
+  name                       = replace("${var.deploy_id}_${var.storage.netapp.volume.name_suffix}", "-", "_")
```
Secretions, line 157: Should we do a regex substitute instead replacing `[^a-z_]`?

Secretions (changes requested):
Only blocking because of automerge

modules/infra/submodules/storage/netapp.tf
miguelhar: sure

Secretions (approved):

Secretions (changes requested):
In before my suggestion goes too quickly

modules/infra/submodules/storage/netapp.tf
Secretions: Sorry, I should have given a full alphanumeric for that: `[^a-zA-z0-9_]`

Secretions (approved):

PR #288, opened 2024-11-01
[DOM-59987] Update linux_user_data.tpl for singe-node based on AL2023

urianchang:
One of the changes from Amazon Linux 2 to Amazon Linux 2023 is switching to `nodeadm` for the initialization process ([docs](https://docs.aws.amazon.com/eks/latest/userguide/al2023.html)). The `single-node` option sets some labels, so we're updating the `linux_user_data.tpl` to be in YAML format.


ddl-dclegg (approved):

urianchang:
Timing this merge with downstream changes is a lot harder than expected.

PR #289, opened 2024-11-06
Update to eks 1.31

ddl-dclegg:


.circleci/config.yml
```diff
@@ -3,13 +3,13 @@ version: 2.1
 parameters:
   helm_version:
     type: string
-    default: "v3.11.2"
+    default: "v3.12.0"
```
Secretions, line 6: Not sure if you want to match what we're actually using, but atm platform-apps is pointing to helm `v3.15.4`

Secretions (approved):

PR #290, opened 2024-11-07
[DOM-62207] VPC Endpint for ECR  and s3 endpoint prefix list to outputs

msingermann-domino:
Context: 
Customers requested to restrict the outbound connection of the user workloads so they can only access the requested IPs
Added a list of allowed CIDRs to apply to the network policies that allows the egress for the services.
Separating the CIDRs in blocks (newrelic, cluster, s3, custom) allows us to configure different blocks to the services that needs them.
In order to get the right CIDRs for the cluster I added those CIDRs to the outputs 
Also as we need to restrict image builder i aded an VPC Endpoint for ECR so we get the IPs for it to add to the allowlist.


modules/infra/submodules/network/vpc.tf
```diff
@@ -49,6 +49,57 @@ resource "aws_vpc_endpoint" "s3" {
...
+
+  lifecycle {
+    create_before_destroy = true
+  }
+  tags = {
+    "Name" = "${var.deploy_id}-ecr"
+  }
+}
+
+resource "aws_vpc_endpoint" "ecr_dkr" {
```
Secretions, line 69: Should we also make these generally optional? That way we/customers don't need to pay for vpc endpoints when the outbound restriction isn't being used?

msingermann-domino: Done!
added a `create_ecr_endpoint` boolean flag in the network section that enables the feature

examples/tfvars/ecr_endpoint.tfvars
```diff
@@ -0,0 +1,19 @@
...
+    availability_zone_ids = ["usw2-az1", "usw2-az2"]
+  }
+  platform = {
+    "availability_zone_ids" = ["usw2-az1", "usw2-az2"]
+  }
+}
+
+network = {
+  create_ecr_endpoint = true
+}
```
Secretions, line 19: nit: no EOF

modules/infra/submodules/network/vpc.tf
Secretions: Cool. We will need to make sure the platform-apps side makes use of this toggle as well.

Secretions (approved):

PR #292, opened 2024-11-12
Miguelhar.ocr

miguelhar:
Adds `modules/capacity-reservation`, used for [PLAT-8864](https://dominodatalab.atlassian.net/browse/PLAT-8864)

### DIFF
--- a/modules/capacity-reservation/main.tf
+++ b/modules/capacity-reservation/main.tf
@@ -1,0 +1,67 @@
+locals {
+  capacity_reservations = merge(flatten([
+    for ng_name, ng in var.instance_capacity : [
+      for az_id in ng.availability_zone_ids : {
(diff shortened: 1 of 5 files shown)


modules/capacity-reservation/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  filter {
+    name   = "instance-type"
+    values = [each.key]
+  }
+
+  location_type = "availability-zone"
+}
+
+
+resource "aws_ec2_capacity_reservation" "this" {
```
ddl-dclegg, line 38: I didn't log this as a change request, but I'm not 100% behind using "this" as an identifier for resources. I understand some things like aws regions, etc., in datasources do that, but it still feels weird and inspecific.

miguelhar: each resource will be identified by the key in the `for_each`

PR #296, opened 2024-11-14
PLAT-8781: Give importer access to flytepropeller

Secretions:
Importer needs access to flyte buckets for lift-and-shift in 6.0.1 too


ddl-kfrench (approved):

Secretions:
Oops, I thought I was doing the "merge when tests finish" thing, but apparently because there's no branch protection, it was actually defaulting to squash. Will keep an eye on the tests and respond appropriately.

Secretions:
Well, the branch not existing caused a failure in and of itself. Wee. Prior commit worked fine and there's no terraform changes since then, so I'm just calling that good.

PR #297, opened 2024-11-14
ensure secrets are deleted

miguelhar:
[PLAT-8971](https://dominodatalab.atlassian.net/browse/PLAT-8971)
There are sporadic instances where the secrets are not being deleted. This change is to ensure they are deleted

### DIFF
--- a/modules/infra/submodules/storage/netapp.tf
+++ b/modules/infra/submodules/storage/netapp.tf
@@ -33,6 +33,8 @@ locals {
     filesystem = "fsxadmin"
     svm        = "vsadmin"
   } : {}
+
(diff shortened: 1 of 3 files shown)


modules/infra/submodules/storage/netapp.tf
```diff
@@ -44,15 +46,78 @@ resource "random_password" "netapp" {
   min_lower   = 1
 }
 
+
+resource "null_resource" "secrets_cleanup" {
```
ddl-kfrench, line 50: we might want to use `terraform_data` since `null_resource` is deprecated in tf >= `1.4` 

https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource

https://developer.hashicorp.com/terraform/language/resources/terraform-data

Secretions (approved):

ddl-kfrench (approved):

PR #319, opened 2025-01-15
feat: enable specifying irsa or role permissions for flyte module

gusmann:
Hi Domino team,

The flyte submodule seems to expect IRSA to be configured on the target EKS cluster. While recommended, there are cases where IRSA isn't installed on an EKS cluster (i.e. following the default Domino EKS installation instructions). This PR allows a user to specify a flag that enables IRSA (disabled by default) and, if disabled, a role to use for attaching the required Flyte S3 policy. If the flag is enabled, instead of requiring the user to specify details about the target cluster, it simply uses the EKS cluster name to pull relevant details from AWS such as the OIDC provider URL/ARN (these can be computed based on the cluster's OIDC endpoint). 

The result is less required variables and support for clusters that do not have EKS IRSA configured.

### DIFF
--- a/modules/flyte/README.md
+++ b/modules/flyte/README.md
@@ -23,20 +23,25 @@ No modules.
 
 | Name | Type |
 |------|------|
+| [aws_iam_policy.flyte_combined](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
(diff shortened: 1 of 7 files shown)


miguelhar:
@gusmann Can you elaborate on what you mean with 
`There are cases where IRSA isn't installed on an EKS cluster`
specifically what is the missing software.

PR #328, opened 2025-01-29
Default karpenter_node_groups to null

miguelhar:


ddl-dclegg (approved):
Looks good to me, it should work fine with the checks in `nodes/main.tf`

PR #329, opened 2025-01-31
PLAT-9123: Costs module non-kms mode

Secretions:
The costs terraform submodule, with cost reporting on, does not currently work when KMS is disabled.

This fixes that so what can use SSE uses SSE, and what can't is simply not encrypted. If you have an issue with the lack of encryption for those things, you should enable KMS, as those two things simply do not support encryption any other way.

### DIFF
--- a/modules/iam-bootstrap/bootstrap-2.json
+++ b/modules/iam-bootstrap/bootstrap-2.json
@@ -43,6 +43,7 @@
         "lambda:GetFunctionCodeSigningConfig",
         "lambda:DeleteCodeSigningConfig",
         "lambda:TagResource",
+        "lambda:UntagResource",
         "lambda:ListTags"
       ],
       "Resource": "*"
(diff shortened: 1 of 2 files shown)


modules/iam-bootstrap/bootstrap-2.json
```diff
@@ -42,7 +41,9 @@
         "lambda:UpdateCodeSigningConfig",
         "lambda:GetCodeSigningConfig",
         "lambda:GetFunctionCodeSigningConfig",
-        "lambda:DeleteCodeSigningConfig"
+        "lambda:DeleteCodeSigningConfig",
+        "lambda:TagResource",
```
ddl-dclegg, line 45: I know it may sound silly, but I've had a couple of errors pop up due to missing the `Untag` correlary, so it might be worth adding `lambda:UntagResource` and (below) `cur:UntagResource`

ddl-dclegg (approved):
My request is non-blocking, but worth consideration

modules/iam-bootstrap/bootstrap-2.json
Secretions: Sure, I'll add that in.

ddl-dclegg (approved):

PR #330, opened 2025-02-04
Fix CIDR outputs for vpc and pod subnets

miguelhar:
The `vpc_cidrs` and `pod_cidrs`  being passed through to the outputs which is inaccurate for byo network.

### DIFF
--- a/modules/infra/submodules/network/outputs.tf
+++ b/modules/infra/submodules/network/outputs.tf
@@ -14,8 +14,8 @@ output "info" {
       pod     = local.pod_route_table_ids
     }
     eips      = [for k, eip in aws_eip.public : eip.public_ip]
-    vpc_cidrs = var.network.cidrs.vpc
-    pod_cidrs = var.network.cidrs.pod
+    vpc_cidrs = local.create_vpc ? aws_vpc.this[0].cidr_block : data.aws_vpc.provided[0].cidr_block
+    pod_cidrs = local.pod_cidr_blocks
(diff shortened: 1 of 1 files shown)


modules/infra/submodules/network/outputs.tf
```diff
@@ -14,8 +14,8 @@ output "info" {
       pod     = local.pod_route_table_ids
     }
     eips      = [for k, eip in aws_eip.public : eip.public_ip]
-    vpc_cidrs = var.network.cidrs.vpc
-    pod_cidrs = var.network.cidrs.pod
+    vpc_cidrs = local.create_vpc ? aws_vpc.this[0].cidr_block : data.aws_vpc.provided[0].cidr_block
+    pod_cidrs = local.pod_cidr_blocks
```
ddl-dclegg, line 18: I'm seeing that we're not setting `local.pod_cidrs` [conditionally](https://github.com/dominodatalab/terraform-aws-eks/blob/f64dd28d46ebe8bec100f22405de2296b6b1ba73/modules/infra/submodules/network/main.tf#L40-L44) the same as vpc & s3. Should we be testing on `local.create_vpc` for `pod_cidrs` as well?

ddl-dclegg (approved):
Looks fine, but I do have one question.

modules/infra/submodules/network/outputs.tf
miguelhar: its indirectly being handled by 
https://github.com/dominodatalab/terraform-aws-eks/blob/main/modules/infra/submodules/network/main.tf#L40

ddl-dclegg:
:shipit:

PR #331, opened 2025-02-12
[DOM-62849] Add S3 Private endpoint Interface

msingermann-domino:
Per [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html) and confirmed by our experimentation as well, our current setup for S3 connectivity is not ‘full privatelink’ (whatever that really means to AWS).

While connectivity is still secure and does not leave the AWS network infrastructure, to be fully 'privatelink' compliant (as to what our customers may expect from that) we are missing to deploy an interface endpoint with private DNS enabled on each deployment.


ddl-dclegg (commented):
Looks good generally, but let's resolve the endpoint [error](https://app.circleci.com/pipelines/github/dominodatalab/terraform-aws-eks/8801/workflows/3f01ed84-89fb-419a-9556-c5fccb58209f/jobs/6136?invite=true#step-115-195677_0)

modules/infra/submodules/network/vpc.tf
```diff
@@ -49,6 +49,39 @@ resource "aws_vpc_endpoint" "s3" {
...
+  lifecycle {
+    create_before_destroy = true
+  }
+  tags = {
+    "Name" = "${var.deploy_id}-s3-endpoint"
+  }
+}
+
+resource "aws_vpc_endpoint" "s3_interface" {
+  count               = local.create_vpc ? 1 : 0
```
Secretions, line 67: Similar to the ecr endpoint, we should have an option to toggle creating this, as this is an interface endpoint, which incurs additional cost (unlike the existing gateway s3 endpoint, which is free).

Secretions (changes requested):

modules/infra/submodules/network/vpc.tf
msingermann-domino: Done!

modules/eks/variables.tf
```diff
@@ -53,6 +56,9 @@ variable "network_info" {
     ecr_endpoint = optional(object({
       security_group_id = optional(string, null)
     }), null)
+    s3_endpoint = optional(object({
```
ddl-dclegg, line 59: This isn't a big deal at all, but this is one place we use `endpoint` and elsewhere it's `interface`. It may make sense to unify on one term in some future update.

ddl-dclegg (commented):
Still looks good to me, will hold my 👍 for @Secretions to determine if his concerns are resolved.

Secretions (approved):

PR #335, opened 2025-02-19
move the security group

miguelhar:


ddl-dclegg (approved):

ddl-dclegg:
@miguelhar Looks like we need to rerun terraform docs

PR #340, opened 2025-02-25
[DOM-65702] Adding support for vpc endpoint services in multiple regions

ldebello-ddl:
Currently private link for nexus only works in the same region, AWS on November release support for cross region vpc endpoint services, the idea is to extend TF config + deployer + infra manifest to support this, it will required for managed data planes ([src](https://dominodatalab.atlassian.net/browse/DOM-65702))


ddl-dclegg (commented):
We just need to bump the provider to `v5.88.0` to handle [supported_regions](https://github.com/hashicorp/terraform-provider-aws/commit/21da8a73bdbcce7bdc9aa702dac68c18b821f639#diff-ccc1fcb04907a24915edca8b56a95f9843cbb8ba37c9c28aaace129ee71c97f0R141)

ddl-dclegg (approved):

PR #344, opened 2025-03-04
correct validation for pod and VPC CIDR network bits

ddl-OBenLagha:


ddl-dclegg:
@miguelhar Can you check this out, where you updated these checks last week?

miguelhar (approved):

PR #356, opened 2025-04-15
allow for predefined irsa policies

miguelhar:
```hcl
variable "irsa_policies" {
  description = "Mappings for custom IRSA configurations."
  type = list(object({
    name                = string
    namespace           = string
    serviceaccount_name = string
    policy              = optional(string) #json
    pod_identity        = optional(bool, false)
  }))
  ```
  
  If you omit `policy` it will try to find it in `./modules/irsa/apps-policies/` which allows for predefined policies as well the case where one wants to overwrite the policy or just not use a predefined.
  
when  `pod_identity = true` it will create the  `aws_eks_pod_identity_association` and a different trust policy for the role

roles:
```bash
./tf.sh cluster output_json irsa | jq .irsa.value.roles
```


modules/irsa/apps-policies/mlflow.json
```diff
@@ -0,0 +1,37 @@
...
+                "s3:DeleteObject",
+                "s3:AbortMultipartUpload"
+            ],
+            "Effect": "Allow",
+            "Resource": [
+                "arn:aws:s3:::${deploy_id}-blobs/*"
+            ]
+        }
+    ]
+}
```
miguelhar, line 37: add new line

modules/irsa/apps-policies/fluentd.json
```diff
@@ -0,0 +1,37 @@
...
+                "s3:DeleteObject",
+                "s3:AbortMultipartUpload"
+            ],
+            "Effect": "Allow",
+            "Resource": [
+                "arn:aws:s3:::${deploy_id}-logs/*"
+            ]
+        }
+    ]
+}
```
miguelhar, line 37: add new line

PR #357, opened 2025-05-09
PLAT-9384 - Create toggle to disable the creation of the oidc provider

miguelhar:
[PLAT-9384](https://dominodatalab.atlassian.net/browse/PLAT-9384)
Allows for a user to opt out of the creation of the EKS cluster OIDC provider as well as to pass their own(`examples/oidc-provider`).


modules/eks/iam.tf
```diff
@@ -277,7 +277,7 @@ resource "aws_eks_identity_provider_config" "this" {
     groups_claim                  = lookup(each.value, "groups_claim", null)
     groups_prefix                 = lookup(each.value, "groups_prefix", null)
     identity_provider_config_name = each.value.identity_provider_config_name
-    issuer_url                    = try(each.value.issuer_url, aws_iam_openid_connect_provider.oidc_provider.url)
```
Secretions, line 280: Is the `identity_providers` value `issuer_url` ignored now?

miguelhar: No,good catch

Secretions (approved):

PR #358, opened 2025-05-09
Set the provider for external dns explicitly

miguelhar:
[PLAT-9383](https://dominodatalab.atlassian.net/browse/PLAT-9383)

Explicitly pass the `aws.global` provider, leveraged by `module.irsa.external_dns`


ddl-dclegg (approved):
I think this looks fine, I'm just wondering why we are shortening to `odc` some places and using `oidc` elsewhere. For example, the flip-flop in `external_dns_oidc_provider_url = var.external_dns.use_cluster_odc_idp ? local.oidc_provider_url : aws_iam_openid_connect_provider.this[0].url`

miguelhar:
> @miguelhar I think this looks fine, I'm just wondering why we are shortening to `odc` some places and using `oidc` elsewhere. For example, the flip-flop in `external_dns_oidc_provider_url = var.external_dns.use_cluster_odc_idp ? local.oidc_provider_url : aws_iam_openid_connect_provider.this[0].url`

that is an old `fat finger` mistake that has carried on.

PR #367, opened 2025-06-11
Add an "extra args" option to our terraform

Secretions:


modules/eks/submodules/k8s/variables.tf
```diff
@@ -24,6 +25,11 @@ variable "ssh_key" {
   })
 }
 
+variable "ssh_extra_args" {
+  description = "Extra arguments passed to SSH"
+  type        = string
```
ddl-dclegg, line 30: Should this also have a `""` default, like in `modules/eks/variables.tf`?

PR #380, opened 2025-07-18
[OT-3419] Adding missing sg rule for public elbs without global accelerator

ldebello-ddl:


modules/load-balancers/security_group.tf
```diff
@@ -22,6 +22,18 @@ resource "aws_security_group_rule" "lb_ingress_from_global_accelerator" {
...
 
+resource "aws_security_group_rule" "lb_ingress_from_public_listeners_without_ddos_protection" {
+  for_each = local.public_listeners_without_ddos_protection
+
+  type              = "ingress"
+  from_port         = each.value.port
+  to_port           = each.value.port
+  protocol          = "tcp"
+  security_group_id = aws_security_group.lb_security_groups[each.value.lb_name].id
+  cidr_blocks       = ["0.0.0.0/0"]
```
miguelhar, line 33: Open to the world, should this be configurable?

miguelhar (approved):

modules/load-balancers/security_group.tf
ldebello-ddl: these are the public elb where before they were created from service annotation without any security group

PR #381, opened 2025-07-21
Fix infra monitoring bucket output, adjusted aws_lb logging to be more deterministic

miguelhar:
* `monitoring_bucket` should not be conditional on `storage.s3.create`, This is fixed in the storage module and causes an issue only when consumed from `infra.monitoring_bucket` works when `infra.storage.s3.buckets.monitoring.bucket_name`
* Create the logging blocks conditionally based on the `enabled` value.
* Currently the logging variables are expecting the `s3_bucket` even when `enabled=False`, make it optional and validate they are passed when `enabled`


modules/load-balancers/main.tf
```diff
@@ -72,16 +72,22 @@ resource "aws_lb" "load_balancers" {
...
+      prefix  = var.access_logs.s3_prefix
+    }
   }
 
-  connection_logs {
-    enabled = var.connection_logs.enabled
-    bucket  = var.connection_logs.s3_bucket
-    prefix  = var.connection_logs.s3_prefix
+  dynamic "connection_logs" {
+    for_each = var.connection_logs.enabled && each.value.type == "application" ? [1] : []
```
ddl-dclegg, line 85: Same as above

PR #386, opened 2025-08-02
updated readme and sh file

ashok-dominodatalab:


ddl-dclegg (changes requested):
I think my only _firm_ ask is to add description to the PR about the reason for the change, where we could also address why we're removing limits.

PR #396, opened 2025-09-10
Custom kms

ldebello-ddl:


modules/eks/variables.tf
```diff
@@ -117,11 +117,15 @@ variable "kms_info" {
...
+    provided_key = If KMS Key was provided
   EOF
   type = object({
-    key_id  = string
-    key_arn = string
-    enabled = bool
+    key_id         = optional(string, null)
+    key_arn        = optional(string, null)
+    enabled        = optional(bool, true)
+    key_policy_arn = optional(string, null)
```
ddl-dclegg, line 127: `key_policy_arn` being null could be problematic. While `null` might be the right default, I think it might make sense to add a `validation` block with an error if it's `null` here, or a lifecycle precondition in the resources where it's used, like role policy attachments

```hcl
lifecycle {
    precondition {
      condition     = local.kms_key_policy_arn != null && local.kms_key_policy_arn != ""
      error_message = "kms_key_policy_arn must be set when provided_key is true."
    }
  
```

modules/infra/iam.tf
```diff
@@ -81,3 +81,9 @@ resource "time_sleep" "create_eks_role_30_seconds" {
   create_duration = "30s"
   depends_on      = [aws_iam_role_policy_attachment.create_eks_role]
 }
+
+resource "aws_iam_role_policy_attachment" "attach_provided_key_policy_to_create_eks_role" {
+  count      = local.provided_key
+  role       = aws_iam_role.create_eks_role.name
+  policy_arn = aws_iam_policy.provided_key_policy[0].arn
+}
```
ddl-dclegg, line 89: nit: missing a newline at the end of the file

modules/infra/kms.tf
```diff
@@ -109,3 +109,18 @@ data "aws_kms_key" "key" {
...
+      kms_key_arn = data.aws_kms_key.key[0].arn
+    })
+  ]
+}
+
+resource "aws_iam_policy" "provided_key_policy" {
+  count  = local.provided_key
+  name   = "${var.deploy_id}-provided-key-policy"
+  policy = data.aws_iam_policy_document.provided_key_policy_document[0].json
+}
```
ddl-dclegg, line 126: nit: missing a newline at the end of the file

ddl-dclegg (approved):

ddl-dclegg (approved):
I won't block on my comments

PR #400, opened 2025-09-23
Workspace Audit buckets, init

miguelhar:
https://dominodatalab.atlassian.net/browse/DOM-71532
Adds Workspace audit buckets


examples/tfvars/workspace-audit.tfvars
```diff
@@ -0,0 +1,28 @@
...
+
+default_node_groups = {
+  compute = {
+    availability_zone_ids = ["usw2-az1", "usw2-az2"]
+  }
+  gpu = {
+    availability_zone_ids = ["usw2-az1", "usw2-az2"]
+  }
+  platform = {
+    "availability_zone_ids" = ["usw2-az1", "usw2-az2"]
```
ddl-dclegg, line 17: Unless it was a purposeful difference, we should remove these quotes around `availability_zone_ids`

ddl-dclegg (approved):
Where it's in the examples, I won't block on it, but I think my only real concern is the quoted [key](https://github.com/dominodatalab/terraform-aws-eks/pull/400/files#r2408040317)

Secretions (approved):

examples/tfvars/workspace-audit.tfvars
miguelhar: Just noticed it is done that way everywhere

PR #421, opened 2025-12-04
DOM-73312 chore(cost-usage-report): Update Athena lambdas to a newer node version

ddl-martin-currao:
## Summary

[DOM-73312](https://dominodatalab.atlassian.net/browse/DOM-73312) — AWS retired the `nodejs16.x` Lambda runtime, so the cost-usage-report Lambdas need to be moved off it. This PR bumps them to `nodejs24.x`, migrates from AWS SDK v2 to v3, rewrites the handler as `async`/`await` (callbacks are not supported on Node 24), and removes scaffolding from the original AWS CloudFormation template that has no purpose under Terraform.

## Changes

**Runtime + SDK upgrade (`templates/aws_cur_initializer.js.tftpl`, `lambda.tf`)**
- Lambda runtime: `nodejs16.x` → `nodejs24.x`.
- Replaced `aws-sdk` (v2) with `@aws-sdk/client-glue` (v3). SDK v3 is bundled in the Node 18+ Lambda runtime, so no `npm`/`node_modules` packaging is needed and no build pipeline is introduced.
- Rewrote the handler as `async`/`await` — Node 24 no longer supports callback-style handlers (`Runtime.CallbackHandlerDeprecated`).
- Fixed the v3-incompatible error path: replaced v2's `JSON.parse(this.httpResponse.body)` + `responseData['__type']` (which silently throws under v3 because `this.httpResponse` doesn't exist) with `err.name === 'CrawlerRunningException'`. The "crawler is already running, treat as success" branch now actually fires under load.
- Moved `GlueClient` instantiation to module scope so it survives warm starts.

**Removed dead CloudFormation custom-resource scaffolding**
- Deleted `templates/cfn-response.js.tftpl` and the `event.ResponseURL` branches from the handler. They were inherited from the AWS-provided `crawler-cfn.yml` (where the Lambda is invoked once as a `Custom::AWSStartCURCrawler` custom resource), but in this Terraform module the Lambda is only invoked by S3 `ObjectCreated` events — the custom-resource path is unreachable.
- Deleted `aws_lambda_function.aws_s3_cur_notification`, its `aws_cur_lambda_executor` IAM role/policy/attachment, the archive_file, and `templates/aws_s3_cur_notification.js.tftpl`. Nothing in the module ever invoked this Lambda — its job (installing the S3 → Lambda notification on the CUR bucket) is done directly by `aws_s3_bucket_notification.aws_put_s3_cur_notification`. Confirmed against Kubecost's own reference template, which also ships this Lambda dead.
- Removed the `CRAWLER_NAME` environment variable on `cur_lambda_initializer` — it has been set but never read since the module's first commit (DOM-49048); the JS hardcodes the crawler name via `templatefile` like `glue_dns` and `glue_region`.

**No changes to**
- The CUR report definition, S3 buckets, Glue database/crawler/status table, Athena workgroup, S3 bucket notification, IAM policy attached to node roles for kubecost queries, the shared lambda security group, SQS DLQ, signing profile, or code-signing config. Kubecost integration outputs (`info.athena_*`, `info.glue_catalog_*`, `info.cur_iam_policy_arn`) are unchanged.

## Migration notes

- **Existing clusters**: `cur_lambda_initializer` updates in place (runtime + code). The dead `aws_s3_cur_notification` Lambda and its IAM stack are cleanly destroyed. Code-signing config has `untrusted_artifact_on_deployment = "Warn"`, so the unsigned zip from `archive_file` deploys without intervention.
- **Clean accounts**: First `terraform apply` provisions everything; AWS Billing's first CUR delivery may take up to 24 h, during which Athena queries return empty (this is an AWS-side limit unrelated to the change).
- **No npm/node tooling**: The Lambda zip is built by Terraform's `archive_file` from the rendered `.tftpl` — `terraform apply` is sufficient.

## Test plan

- [x] `terraform plan` on a clean account shows the expected resource creations (Lambda on `nodejs24.x`, single archive_file, no `aws_cur_lambda_executor` role, no `aws_s3_cur_notification` function).
- [x] `terraform plan` on a cluster currently running the Node 16 version shows: in-place update of `cur_lambda_initializer` runtime+code, destruction of `aws_s3_cur_notification` + its IAM stack, removal of `CRAWLER_NAME` env var, no replacement of crawler/database/workgroup/buckets.
- [x] `terraform apply` succeeds on both scenarios above.
- [x] After apply, drop a CUR parquet object into the bucket prefix (or wait for AWS Billing's natural delivery) and confirm CloudWatch logs show the Lambda invoking `StartCrawler` and the Glue table partitions refresh.
- [ ] Trigger the Lambda twice in quick succession (or while a crawl is mid-flight) and confirm the second invocation logs that the `CrawlerRunningException` was handled gracefully (no Lambda error metric, no DLQ message).
- [x] On a Kubecost-enabled cluster, confirm the cost-analyzer cloud-cost pod queries Athena successfully and the CloudCost UI populates after a CUR file is delivered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ddl-martin-currao:
## Note on the Lambda code-signing setup (out of scope, just documenting)

While walking through `lambda.tf` for this PR I want to flag the current code-signing setup, since it's easy to mistake for actual signing and a future reader may rely on it. **It is currently theatre — the deployed Lambda is unsigned every time.** Spelling out what's there:

- `aws_signer_signing_profile.signing_profile` — creates an AWS Signer profile (`AWSLambda-SHA384-ECDSA`).
- `aws_signer_signing_profile_permission.sp_permission_{start,get}` — grants the account `signer:StartSigningJob` and `signer:GetSigningProfile` on that profile.
- `aws_lambda_code_signing_config.lambda_csc` — a code-signing config that lists the profile as an allowed publisher, with `policies.untrusted_artifact_on_deployment = \"Warn\"`.
- `aws_lambda_function.cur_lambda_initializer.code_signing_config_arn` — binds the CSC to the function.

What's missing for this to actually sign anything: an `aws_signer_signing_job` resource (or an out-of-band signing step in CI) that takes the output of `data.archive_file.cur_initializer_zip` and produces a signed artifact, plus the Lambda pointing at the signed artifact instead of the raw zip. Verified by grep across the repo — no `aws_signer_signing_job`, no `signer start-signing-job` shellout, no external signing pipeline references this profile. The bytes uploaded by the AWS provider are the plain zip from `/tmp/aws_cur_initializer.zip`.

Combined with `untrusted_artifact_on_deployment = \"Warn\"`, the runtime behavior on every apply is:

1. AWS Lambda receives the unsigned zip.
2. The CSC fires its \"untrusted artifact\" check, finds no signature, and emits a warning to CloudWatch.
3. The deploy proceeds anyway.

So the entire signing apparatus exists, costs nothing functional, and produces no real attestation. My read is it's there to make compliance scanners happy (\"a code-signing config is present on the Lambda\") rather than to enforce signature validation. Two ways to make it real, in case anyone wants to follow up later:

- **Enforce + actually sign**: add an `aws_signer_signing_job` (or CI step) that signs `cur_initializer_zip`, point the Lambda at the signed artifact, and switch the CSC to `untrusted_artifact_on_deployment = \"Enforce\"`.
- **Drop it**: remove the four resources above and the `code_signing_config_arn` reference. Net deletion of ~30 lines with no behavior change versus today.

**Not changing anything in this PR** — DOM-73312 is purely the Node 16 → Node 24 / SDK v3 migration plus dead-code removal. Filing this so the next person who looks at `lambda.tf:86–115` knows the signing block is decorative under current usage and can decide whether to harden or remove it.

ddl-martin-currao:
- Tested on mcurra126277 (new cluster)
- Tested upgrade on mcurra126265

Both work. Validated Kubecost pulls data.

ddl-dclegg (approved):
Looks good

PR #443, opened 2026-05-18
DOM-76079: add dns-delegation submodule for cross-cloud NS delegation

miguelhar:
Jira: https://dominodatalab.atlassian.net/browse/DOM-76079

Adds `modules/dns-delegation/`: a small submodule that writes one Route 53 NS record to delegate a subdomain to a list of child-zone nameservers. Built for cross-cloud setups — an AKS dataplane attached to an EKS control plane delegates `<dp>.<cp-root>` to its Azure DNS zone in the same terraform run, by consuming this submodule via `providers = { aws = aws.global }`.

Consumed by cerebrotech/platform-apps#10191.


modules/dns-delegation/variables.tf
ddl-dclegg, line 1: I think adding `allow_overwrite` here would be good, since it's [available](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_record.html#ns-and-soa-record-management) in the [resource](https://github.com/dominodatalab/terraform-aws-eks/pull/443/changes#diff-4f42a260c9914b35abd21defaf483ca376d4a7587450c16aaea043b9bb8c203dR1) (and `terraform import` is a pain sometimes). Defaulting `false` would be fine.

```hcl
variable "allow_overwrite" {
  description = "Allow overwriting an existing NS record with the same name."
  type        = bool
  default     = false
}
```

modules/dns-delegation/variables.tf
```diff
@@ -0,0 +1,20 @@
...
+  type        = string
+}
+
+variable "delegation_name" {
+  description = "FQDN of the subdomain being delegated, e.g. \"azure-east.acme.domino.tech\"."
+  type        = string
+}
+
+variable "nameservers" {
+  description = "NS values from the child zone (Azure DNS, GCP DNS, etc.)."
```
ddl-dclegg: It might be worth noting that Azure exports and [expects](https://learn.microsoft.com/en-us/azure/dns/dns-delegate-domain-azure-dns#:~:text=as%20your%20domain.-,Important,-When%20you%20copy) the trailing dot in a NS while [AWS](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html#:~:text=using%20the%20API.-,Note,-For%20record%20types) allows either. Maybe it's just a note in a doc or comment, or maybe something more, but I saw this and thought it was weird.

ddl-dclegg (approved):
Looks good, just a couple of non-blocking comments.

PR #447, opened 2026-05-28
update node sizing to match our recommended config in the docs

ddl-r-abdulaziz:
According to [documented scale values](https://docs.dominodatalab.com/en/latest/admin_guide/b5da89/set-up-domino-on-eks/#_node_pools), GPU was oversized (1Ti -> 400Gi) and Platform was slightly undersized (100Gi -> 128Gi)


miguelhar:
I would update the gpu node-group and also `modules/infra/variables.tf`

ddl-r-abdulaziz:
@miguelhar done. thanks :D

PR #448, opened 2026-06-09
Adding AmazonEC2ContainerRegistryPullOnly to nodes role

ddl-aabdala:
This PR adds the AmazonEC2ContainerRegistryPullOnly policy to the nodes role to allow the use of ECR Pull-through caches, which specifically require bringing in this particular permission: `ecr:BatchImportUpstreamImage`

The previous policy covering ECR pulls does not include that permission, and in general it is recommended to use the PullOnly variant for this use case but the permissions diff between the two of them is too broad to just swap them out, considering unknown existing use cases that may be relying in these perms. Diff between them:
```diff
$ diff AmazonEC2ContainerRegistryReadOnly AmazonEC2ContainerRegistryPullOnly                                                                 ~/foo 1 ↵
1d0
<                 "ecr:BatchCheckLayerAvailability",
3,5c2
<                 "ecr:DescribeImages",
<                 "ecr:DescribeImageScanFindings"
<                 "ecr:DescribeRepositories",
---
>                 "ecr:BatchImportUpstreamImage"
8,12d4
<                 "ecr:GetLifecyclePolicy",
<                 "ecr:GetLifecyclePolicyPreview",
<                 "ecr:GetRepositoryPolicy",
<                 "ecr:ListImages",
<                 "ecr:ListTagsForResource",
```

### DIFF
--- a/modules/eks/iam.tf
+++ b/modules/eks/iam.tf
@@ -248,6 +248,7 @@ locals {
     "AmazonEKSWorkerNodePolicy",
     "AmazonEKS_CNI_Policy",
     "AmazonEC2ContainerRegistryReadOnly",
+    "AmazonEC2ContainerRegistryPullOnly",
     "AmazonSSMManagedInstanceCore",
     "AmazonElasticFileSystemReadOnlyAccess",
   ]


miguelhar:
this gets you most of the way there, but the first pull of an image that isn't cached yet will still fail. ECR wants `ecr:CreateRepository` from the puller when it has to create the cache repo on the fly, and PullOnly doesn't have it ([docs](https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache-iam.html)). Easiest fix is probably a small statement in custom_eks_node_policy scoped to the cache prefix, e.g. `arn:aws:ecr:*:*:repository/<prefix>/*`. Granting it through the registry permissions policy works too if we'd rather keep it off the node role.

miguelhar (approved):

ddl-dclegg (approved):

PR #449, opened 2026-06-12
PLAT-9951: Update helm to v4, turn off server side apply

Secretions:


.circleci/config.yml
```diff
@@ -290,13 +290,15 @@ jobs:
       - install_tf:
           terraform_version: << parameters.terraform_version >>
       - install_helm:
-          helm_version: << parameters.helm_version >>
+          helm_version: v3.19.2
```
Secretions, line 293: I will do a follow up change that reverts the helm_version change, since this is only needed for CI when the previous version had helm v3.

miguelhar (approved):

